PR #15, opened 2016-04-21
addReduxStore returns store

GantMan:
When I originally read the docs, I did not think `addReduxStore` was modifying `store` which is a const.  Most people will probably see a const as non-changing, even though we know in many ways it can, _like here_.

With this PR people don't have to know.  If they `return store` or if they return `return Reactotron.addReduxStore(store)` both will work.   Dynamic syntax to allow for interpretive implementation.


skellock:
Thanks dude.  

I plan on removing the need for addReduxStore as I should be able to piggyback on the middleware setup.

PR #20, opened 2016-04-21
Opinions and Edits

GantMan:
Removing "Great For" on something required.  It sends a mixed message for newbs.  Changed up flow for "How to Use" section.  I feel this reads better for me.

Also, includes the secrets of `-`, `.`, and `backspace` in readme.


skellock:
Ha.  My precious secrets..... no.....

PR #22, opened 2016-04-22
Add import statement to README example

darinwilson:
Minor tweak to one of the examples in the README. I probably should have known to do that `import` without the docs in the example, but I am a big dummy.


skellock:
Thank you!

PR #23, opened 2016-04-22
Added clear and score to the legend, fixes #5 & #6

### LINKED ISSUE #5: `c` to Clear?
`-` is great but sometimes I'm OCD and want to clear everything between tasks or tests.

kevinvangelder:
Also adds `c` to the keys bound to the `content.clear` command.

### DIFF
--- a/server/commands/menus/menuMain.js
+++ b/server/commands/menus/menuMain.js
@@ -5,6 +5,8 @@ const process = (context, action) => {
     name: 'main',
     commands: [
       {key: 'r', name: 'redux', commands: [{type: 'menu.redux'}]},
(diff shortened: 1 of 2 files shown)


skellock:
The problem is that these key commands are global.  

I agree they should be displayed on the screen, I just don't think they belong in every "command" menu.

skellock:
Oh, there's also '.'.  Which is "repeat" last command.  It works for commands that are flagged as "repeatable".  Such as dispatching or querying redux.

kevinvangelder:
Thanks for the feedback. Updated.

skellock:
I'm ok with this for now.  at least they show in the main menu eh?

PR #25, opened 2016-04-22
Add Reactotron.storeEnhancer alternative.

andrewvy:
This removes the need to add the middleware and also manually call `Reactotron.addReduxStore(store)` by using a store enhancer.

So setting up the Redux support should be as easy as:

```js
import { createStore, compose } from 'redux'
import Reactotron from 'reactotron'

const enhancer = compose(
  // If you have other enhancers or middleware..
  Reactotron.storeEnhancer()
)

const store = createStore(
  rootReducer,
  initialState,
  enhancer
)
```

I tried to keep compatibility with the original workflow, but I've split up the methods to be a bit more reusable. Let me know what you think. The performance tracking seems to be off, but I may have done something silly when I tried to clean it up. This is also still my first week learning Redux. 😁

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Reactotron.connect()
 
 ```js
 // wherever you create your Redux store, add the Reactotron middleware:
+
 import Reactotron from 'reactotron'
 
 const store = createStore(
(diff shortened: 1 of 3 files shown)


skellock:
Ok, now that's cool.  I've never created an enhancer before so this is new territory for me too.  :)

I'm not sure I follow what's going on with the start and stop performance middleware.  Lemme switch to your branch and play around a bit.

Thank you kindly!

skellock:
Heyo!   I brought in your changes from another PR as there was a rebase required.  I had to make a very minor change to integrate Gant's master switch functionality.  I didn't want you to have to do it a 2nd time 😄 

Thanks for the contribution.  And I learned some stuff from this too.  💯

PR #30, opened 2016-04-22
Help menu

GantMan:
includes basic ability to messagebox, and now info box.

image

### DIFF
--- a/src/server/commands/menus/menuHelp.js
+++ b/src/server/commands/menus/menuHelp.js
@@ -5,15 +5,15 @@ const process = (context, action) => {
   const messageText = `
 
     {bold}Hotkeys{/bold}
-    ----------------------------
-      {bold}.{/bold} - Repeat last command
-      {bold}-{/bold} - Insert separator
-      {bold}c{/bold} - Clear reactotron
-      {bold}q{/bold} - Quit
+    ---------------------------------
+      {bold}.{/bold}       Repeat last command
+      {bold}-{/bold}       Insert separator
+      {bold}del{/bold}     Clear reactotron
+      {bold}ctrl-c{/bold}  Quit
 
   `
 
-  context.info(' reactotron {blue-fg}help{/} ', messageText)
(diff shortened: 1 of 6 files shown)


src/server/commands/menus/menuHelp.js
```diff
@@ -0,0 +1,24 @@
...
+const process = (context, action) => {
+
+  const messageText = `
+
+    {bold}Hotkeys{/bold}
+    ----------------------------
+      {bold}.{/bold} - Repeat last command
+      {bold}-{/bold} - Insert separator
+      {bold}c{/bold} - Clear reactotron
+      {bold}q{/bold} - Quit
```
skellock: Should be `ctrl-c` because `q` is specific to the main menu.

src/server/commands/menus/menuHelp.js
```diff
@@ -0,0 +1,24 @@
...
+
+const process = (context, action) => {
+
+  const messageText = `
+
+    {bold}Hotkeys{/bold}
+    ----------------------------
+      {bold}.{/bold} - Repeat last command
+      {bold}-{/bold} - Insert separator
+      {bold}c{/bold} - Clear reactotron
```
skellock: should be `backspace` not `c`.

skellock:
Thanks, this looks really nice.  Good choice of background colour.

Two small things:

`backspace` is clear, not `c`
`q` should be `ctrl-c` as the former is specific to the main menu.  (i'd be cool with removing `q` entirely)

skellock:
Also, wonderful decision to make this a modal.  👍

GantMan:
`c` also clears the screen it seems.  I like `c` over `backspace` due to the windows/mac backspace/delete key naming.  You sure you still want `backspace` to replace `c` for clearing?

GantMan:
@skellock  updated 
image

skellock:
Thx, i hear you on `c`.  Sounds like i'm outvoted too.  /cc @kevinvangelder 

If we use `c` as clear in the future, we'll have to ensure no menus use `c` (like redux > subscribe currently does).

PR #32, opened 2016-04-22
Master switch

GantMan:
This has a breaking change, in that connect params are now a user configuration object.

This PR allows `Reactotron.connect({enabled: __DEV__})` for allowing Reactotron to remain benign in shipped builds.


skellock:
Thank you.  I wanted to add some more options at the client level to opt in to things.  This paves the way.

👍

Issue #35, opened 2016-04-23
Show client-identifying info on connection


skellock:
It'd be cool if `reactotron` showed info about the clients upon connecting.

Right now it says "connected", but could we log an object instead:

```js
{
  version: '0.4.0',
  userAgent: '******',
  reactEnv: 'react-native|react-dom|react-native-web|react-native-desktop',
  platform: 'web|ios|android',
  ip: '127.0.0.1'
}
```

I realize some of this information is just not available without native plugins.  But I'd be happy with the first two to start.

skellock:
Today's feature brought to you by the number #37 .

PR #37, opened 2016-04-23
Add multiple client identification and display.

andrewvy:
image

Sets up a new display for the list of clients so we can move forward with multiple client support.

- Adds #35 basic client information, starting with `client.ip`, `client.version` and `client.userAgent`.
- Changes the bottom right 'ONLINE'/'OFFLINE' to a client count!
- Clients are stored in the `context.clients` object, where `key = socket id` and `value = client info`


skellock:
You are awesome.  Will try to get this in tonight.

skellock:
Looks great.  I'm actually going to merge as is.

One thing I'd like to do is promote the `client connection` box to a `message` box off a keystroke.

It's great info, but I don't think it should be on the screen at all times.  Maybe just emit the message in the `log` box upon connection.  Then hit a key stroke to show that info you have in the panel.

I can do that if you'd like though.  :)

andrewvy:
I like either option. 😄 Should be easy to split it out.

I'd be curious about what you think about the multi-client experience. Right now, we get all events for all clients into a single display. Which can be weird if we subscribe to events on the Redux store.

Maybe a client menu w/ client switching functionality? Where you can subscribe to certain redux keys globally, but switch between different clients for their redux store value?

skellock:
Ya it doesn't make a lot sense to have have 3 clients and show everything. As you said, it really breaks apart when dealing with state.

I like your idea of having a client selector to dial into the appropriate channel.

Man.  So many possibilities!

PR #42, opened 2016-04-24
Add client popup menu for list of clients.

andrewvy:
image

This could lead to the client switcher that I mentioned in #37 

Let me know if this is what you expected. I just added this in after your initial feedback. :+1:


andrewvy:
Ack, I should have a newline after the list of clients. One moment.

skellock:
Love it!

PR #59, opened 2016-07-14
Get correctly localhost for Android emulator

jhen0409:
Use `NativeModules.AndroidConstants` to get correctly localhost `10.0.2.2` (Genymotion: `10.0.3.2`) for Android emulator.

It mean we will never need use `add reverse` for Android emulator.


skellock:
Lovely!

PR #66, opened 2016-07-29
Add support of Symbol action

fantasywind:
Hi, I added simple patch to support symbol action in reduxMiddleware


skellock:
Thanks.  I auto-generate that dist file... (I should really just remove it from git!)

skellock:
Thanks for the patch!

PR #73, opened 2016-07-31
Logger be logging

skellock:
image


kevinvangelder:
You could probably make the debug messages blue with an info icon if you wanted to, but not very important.

PR #104, opened 2016-08-12
Adds benchmark rendering.

skellock:
image


maluramichael:
Is it possible to bench with 'reactotron-react-native'? Because 'bench' is an unknown method.

skellock:
Oops.  Missed this comment.

You want to create a `bench` with `.benchmark`.

Check this out:

https://github.com/reactotron/reactotron/blob/master/packages/reactotron-core-client/scripts/send.js#L13

PR #133, opened 2016-08-18
Add help dialog with CMD + ?

GantMan:
Looks/works great
image

### BUT
one has to ask themselves, how would they ever know to press CMD+? in the first place?

I have two answers to that.  1, you'd mention it in the intro.  2, we can have an indicator in the background like Atom does on an empty Timeline.

Thoughts?


GantMan:
## I'm contributing!
iwae1wo - imgur

skellock:
NICE!!  That looks really sharp Gant.

Ya, so the help menu?   That's what the empty space up top is.  I'll merge this in and add the button to kick this off.

Welcome aboard dude!

PR #138, opened 2016-08-20
Renders actions with flat data as a table instead of a tree.

skellock:
image


GantMan:
Good call.  Does your table support tree innards though?   I can see someone passing an object.

skellock:
If a nested object is passed, it'll switch to object mode.  if the object is shallow/flat, it'll pick table rendering.

https://github.com/reactotron/reactotron/blob/next/packages/reactotron-app/App/Lib/IsShallow.js#L4

PR #175, opened 2016-08-25
Fix typo in `reactotron-apisauce` README

patricksmith:
After following the directions in the `reactotron-apisauce` README, I noticed API calls were not being logged. The README says to use `Reactotron.use(tronsauce)` while the demo app uses `Reactotron.use(tronsauce())`. After changing to calling `tronsauce` the calls started showing up in Reactotron.


skellock:
Oops.  Thank you.

Issue #173, opened 2016-08-25
Question about Reactotron and production builds.


avishayil:
Does the logging working only on dev?
I mean, logging code wouldn't run on production and will crash the app, right?
Asking to make sure.

skellock:
There's nothing stopping you from running in production builds at all.  Just be sure to install with `--save` instead of `--save-dev`.

I don't recommend it for shipping to the App Store or a running web site, however, if you're looking to debug some problems with a production build on a local device, it's totally a legit approach.

avishayil:
How can I make sure code-wise that Reactotron is running on DEV only? do I have an option to do that without maintaining two repos?
On the previous Reactotron repo, we had an option:
````
Reactotron.connect({enabled: __DEV__})
````

Is it possible to do it here?

skellock:
What I do is wrap the entire block in `if (__DEV__)`.

skellock:
Another option is just wrapping the `connect()` call in `if (__DEV__)`.

avishayil:
Like this?

````
import Reactotron from 'reactotron-react-native';
import apisaucePlugin from 'reactotron-apisauce';

Reactotron
  .configure() // we can use plugins here -- more on this later
  .use(apisaucePlugin()) // <-- here we go!!!

if (__DEV__) {
  Reactotron
  .connect() // let's connect!
}
````

Then, do I need to wrap every `api.addMonitor(Reactotron.apisauce);` call with `if (__DEV__)` ?

skellock:
Yep, give that a go.

You shouldn't have to wrap all your call sites with this guard though, because all communication from your app back to reactotron  flows through this 1 line of code:

https://github.com/reactotron/reactotron/blob/master/packages/reactotron-core-client/src/index.js#L123

If you never connect, you'll never send anything.  So you can leave your monitors and logging calls in place if you really want to.

avishayil:
By the way - theoretically - why don't you recommend doing it with app-store build? Doesn't it gives you the ability to debug devices without running special build on them?

skellock:
1. It uses websockets.  Which will keep the draining the battery.
2. It's a privacy violation.  If you stored some social tokens in state, we could grab them.
3. A security violation as it opens up a backchannel to control the app.  Haha.

Reactotron is a debugging tool.  At least, that's how I recommend you use it!  =)

avishayil:
Thanks for the clarifications!

lukesneeringer:
[...]

PR #181, opened 2016-08-27
Switches the demo-react-js app to dev dependency only.

skellock:
There was some questions about using Reactotron in `--save-dev` vs. `--save`.  

My advice was to run in `--save-dev` only, but as seen in #173, I didn't quite follow through with "why" OR "how"?  In fact, as @lukesneeringer points out.  My advice leads to a broken build.

I've recently written about [why](https://github.com/reactotron/reactotron/blob/master/docs/tips.md), but this PR shows how.

Again, nothing is stopping you from running `--save`, just please don't run `Reactotron.connect()` in production builds.

Thanks!  Sorry for the confusion.

### DIFF
--- a/packages/demo-react-js/README.md
+++ b/packages/demo-react-js/README.md
@@ -1,4 +1,19 @@
-## Reactotron Notes
+# Reactotron Notes
+
+## Using in dev mode only
+
+I've altered this example to show a few techniques you can use in order to keep Reactotron living as a dev dependency.
+
+You'll notice that we don't `import`, we `require` because you can conditionally `import` in ES6.
+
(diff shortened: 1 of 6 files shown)


avishayil:
Can you please create a similar docs for RN environment?
Thanks for your work!

mcmar:
@skellock I searched around and arrived here because the docs for react-native still recommend using `--save-dev` along with `import Reactotron from 'reactotron-react-native';`. That will break your build, so the docs for react-native are still totally busted. Thought you might want to know.

skellock:
Yeah. That’s the quick start guide but ya we could be more clear about these things.

There is a little more info about this here: https://github.com/infinitered/reactotron/blob/master/docs/tips.md#running-in-production

My preference is to wrap with `__DEV__` guards.  

We are working on slimming it down so it won’t matter so much if you want it in prod too.

PR #194, opened 2016-08-31
added logging tips

GantMan:
Just a few quick tips for docs.  Will make it better next round.  Could use pictures etc.


skellock:
Bonus points for emoji.

PR #217, opened 2016-10-20
Add `npm install` step to Contributing doc

patricksmith:
Running `npm run welcome` without first installing `lerna` will result in a number of errors.

I fixed this by running `npm install` before `npm run welcome`, but I'm not sure if the expectation is that `lerna` be installed globally.


skellock:
Thank you!

PR #218, opened 2016-10-20
Show request params in Reactotron App

patricksmith:
Adds request params to the API responses shown in Reactotron:

screen shot 2016-10-20 at 10 42 59 am

I find myself often writing requests like `api.get('/classes/Config', { limit: 1 })`, and thought it would be useful to view the params in Reactotron.


skellock:
Thank you.

PR #233, opened 2016-11-08
Pass the current state to onRestore as well

rmevans9:
I have a need to leave a part of my state tree untouched when restoring the state. Adding the ability to pull from the current state in the onrestore allows for this use case.


skellock:
Ahhh... perfect!  This gives me some ideas at the UI layer too to help with this use case.

Thanks!

Issue #234, opened 2016-11-08
Redux Saga Monitor


rmevans9:
I am working on starting a React Native app that will rely heavily on Redux Saga and would love to see Reactotron be a bit more involved with testing when it comes to Sagas. I am playing around with the Redux SagaMonitor that is included in the examples folder of the repo and am considering how this might be able to be incorporated into Reactotron. Currently I have it printing to the console which means I have to have Remote Debugging turned on. I feel like this could be a great opportunity for an additional screen in Reactotron.

Ref: https://github.com/yelouafi/redux-saga/blob/master/examples/sagaMonitor/index.js

Thoughts?

skellock:
Absolutely.

I've already started on this too.  There's a few issues with surfacing the information out of redux-saga though.

I plan on asking @yelouafi once I'm able to formulate the right question to ask.

My problem specifically is knowing that a generator is finished.

I know when an effect starts, and there's a `id` that i can use to track back a nested effect.

But if a generator gets kicked off via a `takeLatest()`, that becomes a fork signal on `take` that ends immediately.  Even before the list of sub effects finish.

I get that within a `while(take())`, the function never "finishes", but I feel like in the case of `takeLatest()`, it'd be nice for the logging system to know the triggering generator is done.

With that mechanism in place, I could move on to the visualization part (which I do have a few ideas for).

He's also working on a visualizer (https://github.com/yelouafi/redux-saga/pull/609) too, so maybe he's encountering this too?

Or maybe I'm just holding it wrong?

It's probably the latter.

yelouafi:
@skellock 
When a saga is forked (or spawned) the method `monitor#effectResolved(id, result)` is invoked with a `Task` as the result. [In the monitor example](https://github.com/yelouafi/redux-saga/blob/master/examples/sagaMonitor/index.js#L77-L87) you can see how the end of the task is tracked (via chaining a handler to the `done` promise and reeinvoking resolveEffect again). Similarly there is an [`effectCancelled` method](https://github.com/yelouafi/redux-saga/blob/master/examples/sagaMonitor/index.js#L60-L64) that is invoked when a saga (or any other effect) is cancelled.

Perhaps I dont quite understand the issue with `takeLatest` but as it's just a helper it'll fire the sequence of the individual effects inside it which will trigger a sequence of monitor callbacks (effectTriggered, effectResolved/effectRejected/effectCancelled).

But feel free to ask any question. I'll be glad to help.

skellock:
[...]

PR #235, opened 2016-11-09
Adds redux saga support.

skellock:
Discussed in #234.  

Thanks to @rmevans9 for suggesting the feature.  

Thanks to @yelouafi for guidance on how to build this puppy.

reactotron-redux-saga

There's so much more we can do, but this is a usable first start.


rmevans9:
This is excellent! Can't wait for this to land.

PR #252, opened 2016-11-12
Round the duration for sagas

### LINKED ISSUE #240: Trim the floating point precision you monster
image

rmevans9:
Fixes #240 

I just did a straight round on the client side. I considered doing it on the app side but wasn't sure which would feel more "right" so just went with client side.


skellock:
Ya thats totally fine.

PR #256, opened 2016-11-15
Fix typo in saga effects title

patricksmith:
Just a quick typo fix -- previously, if there were 0 effects, the title would be rendered as "0 Effectfalse":

image

 Instead, only omit the "s" in "effects" when there is exactly 1 effect.


skellock:
Thank you! Oopsie.

PR #260, opened 2016-11-15
A quick fix to not crash with the new Redux Saga

### LINKED ISSUE #259: Redux Saga 0.13.0 + Reactotron Saga Monitor = SADNESS
It looks like 0.13.0 added a new call on the saga monitor that isn't being passed in by the reactotron saga monitor. We end up with:

`sagaMonitor.actionDispatched is not a function`

from Redux Sagas release notes:

> added actionDispatched trigger to the monitor contract. We need this in order to figure out saga/take effects that reacted to a given action. Also added a silent flag SAGA_ACTION (non enumerable property) added to distinguish actions dispatched by sagas from others.

A quick fix might be to just pass a noop function until it is figured out if this new monitor function could help

rmevans9:
Fixes #259

### DIFF
--- a/packages/reactotron-redux-saga/src/saga-monitor.js
+++ b/packages/reactotron-redux-saga/src/saga-monitor.js
@@ -215,6 +215,7 @@ export default (reactotron, options) => {
     effectTriggered,
     effectResolved,
     effectRejected,
-    effectCancelled
+    effectCancelled,
+    actionDispatched: () => {}
   }
 }


rmevans9:
I tested on my production app with the new version of Redux Saga but didn't update the example apps just yet as I needed a fix ASAP to wrap up some things. I can update the examples probably in the next few days if someone doesn't beat me to it.

skellock:
Thank you for the quick PR.

PR #265, opened 2016-11-16
Correct devtool type 'source-map' instead of 'source-maps'

nhducit:
Correct devtool type 'source-map' instead of 'source-maps'


skellock:
Oops.  Thanks man.

PR #267, opened 2016-11-17
Default the trigger type

rmevans9:
The trigger type showed up in the command the way I was hoping so I THINK that is where it belongs. If not I can move it around.


skellock:
Yep.  That's a fine fallback.  Thank you sir.

Issue #276, opened 2016-11-23
Read source-maps to display exceptions more userfriendly


reicheltp:
Currently, if I log an error reactotron just display the plain error object with reference to the error line in source code. Unfortunately, the source is packed and this info is useless for the developer. 

![image](https://cloud.githubusercontent.com/assets/4380869/20578209/7313c33c-b1c6-11e6-91de-70717cdc69bf.png)

It would be super awesome if reactotron loads the corresponding source-maps and displays the real error line. Reactotron could also show the corresponding code-snippet, but this is maybe too much.

skellock:
This case should be supported.  Looks like you're doing React Native Android there.

Checkout https://github.com/reactotron/reactotron/blob/master/docs/plugin-track-global-errors.md

There's a error tracking plugin you'll need to setup in your `ReactotronConfig.js` that should capture this.

What you'll have to do is let that error bubble up unhandled in order for React Native's error handling & sourcemap resolving to kick in.

Having said... it'd be awesome to call their sourcemap resolver on demand!  Like if you're in the catch block and you want to continue normal flow.

rmevans9:
@skellock RE: Call their sourcemap resolve on demand - that is a large part of the code in #278 so we could possibly abstract that logic out a bit and make use of it in more places then just the saga logging.

skellock:
This is all in place now.  :1st_place_medal:

PR #278, opened 2016-11-26
Redux Saga Logging Magic

rmevans9:
Note: This has mostly been tested in my own app but will be setting up to test it in the demo app over the next few days.

I had to have the "end user" pull two internal methods and pass them to Reactotron as I was unable to find a way to pull them within the reactotron code without it throwing errors.

This PR should be considered a WIP

### DIFF
--- a/packages/demo-react-native/App/Config/ReactotronConfig.js
+++ b/packages/demo-react-native/App/Config/ReactotronConfig.js
@@ -1,3 +1,5 @@
+import parseErrorStack from 'parseErrorStack'
+import symbolicateStackTrace from 'symbolicateStackTrace'
 import Reactotron, { trackGlobalErrors } from 'reactotron-react-native'
 import tronsauce from 'reactotron-apisauce'
 import { reactotronRedux } from 'reactotron-redux'
@@ -5,7 +7,7 @@ import sagaPlugin from 'reactotron-redux-saga'
 
 if (__DEV__) {
   Reactotron
-    .configure({ name: 'React Native Demo' })
+    .configure({ name: 'React Native Demo', parseErrorStack, symbolicateStackTrace })
     .use(tronsauce())
     .use(reactotronRedux({
       isActionImportant: action => action.type === 'something.important',
--- a/packages/demo-react-native/App/Containers/RootContainer.js
+++ b/packages/demo-react-native/App/Containers/RootContainer.js
@@ -7,6 +7,7 @@ import Repo from '../Components/Repo'
 import { Actions as RepoActions } from '../Redux/RepoRedux'
 import { Actions as LogoActions } from '../Redux/LogoRedux'
 import { Actions as StartupActions } from '../Redux/StartupRedux'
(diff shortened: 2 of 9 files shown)


skellock:
I can't tell you how happy this makes me!  Will check it out this weekend.

rmevans9:
In trying to see if there is a way to not have the error logged the entire way up the saga chain I couldn't find one as each of the sagas threw the error as if the error occurred in their saga. I am thinking we might need to do more filtering based on the fact that if the exact error was thrown already instead of trying to only allow the closest saga to throw it.

skellock:
I put this in:
```js
          // only log the error once
          if (!error.reactotronWasHere) prettyLogSagaError(error)
          error.reactotronWasHere = true
```

Low tech, sure.  It works though.

skellock:
So your code works great.  

I'm thinking we don't pass the RN error functions into saga though.  I think on slack you'd mentioned putting another [function on Reactotron](https://github.com/reactotron/reactotron/blob/master/packages/reactotron-react-native/src/plugins/track-global-errors.js#L71) called `reportError` or something.

Then in our monitor, we just call `Reactotron.reportError && Reactotron.reportError(error)`.

`reportError` is essentially just `prettyLogSagaError()` relocated.  =)

And now we have a way to also report errors needed for #276 .

This'll be the first time two plugins "talk" to each other, but I'm 100% down with that.

rmevans9:
Nice! I was starting to move in that direction and starting to see if the same error object is passed to each of the promises... I see that is the case! Very nice.

skellock:
Ya.  You kinda nailed it here dude.  Great work.

rmevans9:
Yeah, I think moving that logic works. I will work on moving that around. When we setup Reactotron we will need to still pass in those internal RN functions because I still have not found a way to pull them from within the plugin itself. But wrapping the use of them into a more generic method on the Reactotron object I think would be fantastic.

skellock:
Two things I learned here:

* It fires in triplicate because Root Saga -> takeLatest -> Our Running Saga.  That's 3 tasks.  So it pops up the tree rejecting the promises along the way.

* This doesn't catch `put` effects.  And apparently that's on their radar (https://github.com/yelouafi/redux-saga/issues/632).  Apparently `put.sync` (called `put.resolve` next build) is a thing.  I had no idea.  I had thought it was sync (which now makes a tonne of things more clear!)

Since our goal is just monitoring (not control flow), it might be worth seeing of `put`-bombs show up in a few other places:

* the logger on `createSagaMiddleware`
* the `onError` on `createSagaMiddleware`
* or; even some caveman wrapper implementation of `runSaga(_, { dispatch: ... })`

Just spitballing of course.

I find in my journeys, a make many errors on the React components caused by `PUT`s.  I'd love to have that visibility too.

rmevans9:
I am sure one of those options will log put failures. I'll get an example setup and test later today.

rmevans9:
Rearranged code to make this more solid. This still doesn't handle `puts`. That is next on my list. Updated the demo app to show off the neat features

Issue #279, opened 2016-11-27
React-native socket.io-client require problem


ricardofbarros:
Currently after a fresh install I get this `Requiring unknown module "socket.io-client/socket.io". I traced it back to this [PR](https://github.com/socketio/socket.io-client/pull/1025)

They changed the path of the minified file. I will be pushing a PR shortly addressing this changes.

skellock:
Thank you.  I had a lot of issues getting this behaving on both React JS and React Native.  I hadn't realized it was busted.

skellock:
Thanks again.  This is now shipping in `1.5.3`.  Cheers man!

Issue #199, opened 2016-09-08
window is not defined


jeveloper:
Hi there

My app is still fairly barebones

Any hints?

ReferenceError: window is not defined
    at Object.<anonymous> (/node_modules/reactotron-react-js/dist/index.js:95:14)
    at Module._compile (module.js:541:32)
    at Module.require._compile (node_modules/gluestick/node_modules/webpack-isomorphic-tools/babel-transpiled-modules/index.js:414:28)
    at Module._extensions..js (module.js:550:10)

skellock:
I see you're using isomorphic... is perhaps this getting loaded on the server, where `window` isn't a thing?

Does this help https://github.com/reactotron/reactotron/issues/129#issuecomment-240091289?

jeveloper:
You're right
i use gluestick with my reactapp and this is happening on node side during transpiling 

i moved this code to componentDidMount 
this way window is present, but nothing seems to happen except an error is gone.

i'm not using barebones reactapp from fb , mine is with reducers, actions and component structure, curious where the right place for reactron to live in :)

thanks

FYI , brew cask had version 1.1.1 only though

skellock:
Hmm.  With isomorphic, does the redux store live client side or server side?

Also, 1.1.1 is the latest release of the app.  I'm opted to use a shared version number (via lernajs) on all the components for this monorepo.

jeveloper:
It should be on client, but i think the process of babel transpiling is messing this up
ill close the ticket :)

thanks !

messense:
This causes some problem when running test cases using serverless Node environment in react-native. Had to do some dirty hack to workaround it:

```javascript
global.window = {
  navigator: {}
};

skellock:
Interesting, perhaps [this is causing a problem](https://github.com/reactotron/reactotron/blob/master/packages/reactotron-react-native/src/index.js#L28)?

messense:
@skellock Yes. But `window` is `undefined` too when running tests in Nodejs using mocha.

`window` is not defined makes it hard to write testable code.

skellock:
Gotcha.  It sounds as if Reactotron shouldn't be loaded during these tests.  That kinda makes sense.  So perhaps ensuring `__DEV__` is set to false during your tests would be an answer?

Lemme know if there's anything you think I can/should do to help.

RusAlex:
Seems reactotron not well encapsulated it's logic. It mustn't mutate `window` global object during importing. 

Thanks

[...]

PR #280, opened 2016-11-27
Fix socket.io-client require for reactotron-react-native

ricardofbarros:
Address #279.

I didn't bump the versions of the other packages as you did before, because I don't know if it's really necessary.


skellock:
Yep.  I'll bump all the downstream dependencies.  I use [lerna](https://github.com/lerna/lerna) to manage that stuff.

Thank you for bringing this up.

skellock:
Thank you.  This also removes the need to hack in the `window.navigator` in #199 .

PR #299, opened 2016-12-07
Unwrap dev tools action if it looks like a dev tools action

rmevans9:
Not sure if there is a better way to find out if redux dev tools is turned on but here is a quick and dirty "detection" that unwraps it so Reactotron looks right even with redux dev tools on


skellock:
Thank you.

PR #304, opened 2016-12-10
[React Native] Fix auto-detect Android server port for newer RN versions

jhen0409:
Looks like `__fbBatchedBridge.RemoteModules` has been removed on newer versions of RN, so currently `get-host` doesn't work.

A workaround is use `__fbBatchedBridgeConfig.remoteModuleConfig` instead.


skellock:
Thank you kindly!!

Issue #285, opened 2016-11-28
Reactotron app not connecting (cli works fine)


chapati23:
I'd like to give a detailed bug report but tbh I have no idea how to debug the desktop app, it's just not connecting.

I'm using the ignite starter kit, this is my ReactotronConfig:

```js
import Immutable from 'seamless-immutable'
import { StartupTypes } from '../Redux/StartupRedux'
const Reactotron = require('reactotron-react-native').default
const errorPlugin = require('reactotron-react-native').trackGlobalErrors
const apisaucePlugin = require('reactotron-apisauce')
const { reactotronRedux } = require('reactotron-redux')
const sagaPlugin = require('reactotron-redux-saga')

if (__DEV__) {
  Reactotron
    .configure({
      // host: '10.0.3.2' // default is localhost (on android don't forget to `adb reverse tcp:9090 tcp:9090`)
      name: 'Onboarding' // would you like to see your app's name?
    })

    // forward all errors to Reactotron
    .use(errorPlugin({
      // ignore all error frames from react-native (for example)
      veto: (frame) =>
        frame.fileName.indexOf('/node_modules/react-native/') >= 0
    }))

    // register apisauce so we can install a monitor later
    .use(apisaucePlugin())

    // setup the redux integration with Reactotron
    .use(reactotronRedux({
      // you can flag some of your actions as important by returning true here
      isActionImportant: action => action.type === StartupTypes.STARTUP,

      // you can flag to exclude certain types too... especially the chatty ones
      // except: ['EFFECT_TRIGGERED', 'EFFECT_RESOLVED', 'EFFECT_REJECTED', 'persist/REHYDRATE'],

      // Fires when Reactotron uploads a new copy of the state tree.  Since our reducers are
      // immutable with `seamless-immutable`, we ensure we convert to that format.
      onRestore: state => Immutable(state)
    }))

    // register the redux-saga plugin so we can use the monitor in CreateStore.js
    .use(sagaPlugin())

    // let's connect!
    .connect()

  // Let's clear Reactotron on every time we load the app
  Reactotron.clear()

  // Totally hacky, but this allows you to not both importing reactotron-react-native
  // on every file.  This is just DEV mode, so no big deal.
  console.tron = Reactotron
} else {
  // a mock version should you decide to leave console.tron in your codebase
  console.tron = {
    log: () => false,
    warn: () => false,
    error: () => false,
    display: () => false,
    image: () => false
  }
} 
```

chapati23:
[...]

PR #307, opened 2016-12-14
Update config to include connect

haikyuu:
This PR adds `connect` to the redux example. I thought since it has the import statement, it should also have the call to `connect`.
### Story
I did a dumb copy paste when i came here to see how to add the redux plugin.
Then, i spent several dozens of minutes to figure out why reactotron didn't connect #285 
This could hopefully save other people those precious minutes.


skellock:
Oops, Thank you!  Ya, that was my bad.

PR #315, opened 2016-12-26
ImmutableJS Support

rmevans9:
This PR introduces basic ImmutableJS support. Note that it has only been tested with react-boilerplate. I imagine there will be some setups that this doesn't cover but it at least covers a basic one to get us started.


rmevans9:
NOTE: I don't know much about Rambda to know if the getting of the cleaned state could be incorporated right into the usage of it (mainly talking about get-subscription-values). if it can please just direct me to how and I will gladly update this PR

skellock:
I'm ok with this.  If you'd like, you could make it:

```js
import { when, prop } from 'ramda'

const pojoify = when(prop('toJS'), x => x.toJS())
```

But I'm cool with what you did.

skellock:
I won't be able to cut the next build until the 29th (when I'm back home).

Thanks for doing this!

PR #318, opened 2017-01-07
Support `exceptionPatterns` in `reactotron-redux`

g6ling:
This PR adds `exceptionPatterns` which is `exception` for Function. I think this make `except` more simple.

example
```js
const REACT_NATIVE_ROUTER_FLUX = (type) => type.includes('REACT_NATIVE_ROUTER_FLUX');
const REDUX_SAGA_REQUEST = (type) => type.includes('request');

Reactotron
  .use(reactotronRedux({
    except: [],
    exceptPattern: [
      REACT_NATIVE_ROUTER_FLUX,
      REDUX_SAGA_REQUEST,
    ],
  }))
```

### DIFF
--- a/packages/reactotron-redux/src/create-action-tracker.js
+++ b/packages/reactotron-redux/src/create-action-tracker.js
@@ -15,6 +15,7 @@ export default (reactotron, trackerOptions = {}) => {
   // assemble a crack team of options to use
   const options = merge(DEFAULTS, trackerOptions)
   const exceptions = concat([DEFAULT_REPLACER_TYPE], options.except || [])
+  const exceptionPatterns = options.exceptionPatterns || [];
 
   // the store enhancer
   return next => (reducer, initialState, enhancer) => {
@@ -39,8 +40,10 @@ export default (reactotron, trackerOptions = {}) => {
 
         var unwrappedAction = action.type === 'PERFORM_ACTION' && action.action ? action.action : action
 
+        const exceptionPattern = exceptionPatterns.reduce((prev, pattern) => pattern(unwrappedAction.type) || prev, false);
+
         // action not blacklisted?
-        if (!contains(unwrappedAction.type, exceptions)) {
+        if (!exceptionPattern && !contains(unwrappedAction.type, exceptions)) {
           // check if the app considers this important
           let important = false
           if (trackerOptions && typeof trackerOptions.isActionImportant === 'function') {
--- a/packages/reactotron-redux/src/create-action-tracker.js
+++ b/packages/reactotron-redux/src/create-action-tracker.js
@@ -1,4 +1,4 @@
-import { contains, is, concat, merge } from 'ramda'
+import { is, concat, merge } from 'ramda'
 import { DEFAULT_REPLACER_TYPE } from './replacement-reducer'
 
 const DEFAULTS = {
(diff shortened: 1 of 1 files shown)


skellock:
Hey!  Thanks man.

I like this feature.  Now that I see this in action, though, I'm thinking we add this to `except` instead of introducing a new property.  I totally get why you did, and I appreciate it.

Let's make `except` an array that accepts 3 things:

* a string (exact match)
* a regex (to uh... regex match)
* a function (to invoke -- just like you wrote)

That way, we get your new goodies, we get a regex, and we keep the old way, all without breaking the interface!

Thoughts?

g6ling:
Thanks. I update my PR. I think this code will work as you want.
Should I update this(https://github.com/reactotron/reactotron/blob/master/docs/plugin-redux.md)?

packages/reactotron-redux/src/create-action-tracker.js
```diff
@@ -40,10 +39,24 @@ export default (reactotron, trackerOptions = {}) => {
...
+              return exception(actionType);
+            case 'object':
+              return exception.test(actionType);
+              break;
+            default:
+              return false;
+          }
+        }
+
+        const checkExceptions = exceptions.reduce((prev, exception) => checkExcpetion(exception, unwrappedAction.type) || prev, false);
```
skellock: The function `checkExcpetion` might not exist.

packages/reactotron-redux/src/create-action-tracker.js
```diff
@@ -40,10 +39,24 @@ export default (reactotron, trackerOptions = {}) => {
 
         var unwrappedAction = action.type === 'PERFORM_ACTION' && action.action ? action.action : action
 
-        const exceptionPattern = exceptionPatterns.reduce((prev, pattern) => pattern(unwrappedAction.type) || prev, false);
+        const checkException = (exception, actionType) => {
+          switch (typeof exception) {
+            case 'string':
+              return actionType === exception;
```
skellock: We use the `standard` standard.  If you run `npm run lint` you'll see it won't like semi-colons.

packages/reactotron-redux/src/create-action-tracker.js
```diff
@@ -40,10 +39,24 @@ export default (reactotron, trackerOptions = {}) => {
...
+            case 'function':
+              return exception(actionType);
+            case 'object':
```
skellock: Too bad we couldn't detect regexp's!  Oh well.  You win this time JavaScript.

packages/reactotron-redux/src/create-action-tracker.js
```diff
@@ -40,10 +39,24 @@ export default (reactotron, trackerOptions = {}) => {
...
+              return exception.test(actionType);
+              break;
```
skellock: You can nuke this `break`.  The return just above it will take care of this case.

packages/reactotron-redux/src/create-action-tracker.js
```diff
@@ -40,10 +39,24 @@ export default (reactotron, trackerOptions = {}) => {
...
+            default:
+              return false;
+          }
+        }
+
+        const checkExceptions = exceptions.reduce((prev, exception) => checkExcpetion(exception, unwrappedAction.type) || prev, false);
 
         // action not blacklisted?
-        if (!exceptionPattern && !contains(unwrappedAction.type, exceptions)) {
+        if (!checkExcpetions) {
```
skellock: Looks like this variable should be `checkExceptions`?

g6ling:
thanks for your review. How can I test my code in reactotron?

skellock:
```sh
cd packages/reactotron-redux
npm test
```

skellock:
You need any help wrapping this up?

packages/reactotron-redux/src/create-action-tracker.js
```diff
@@ -39,8 +39,24 @@ export default (reactotron, trackerOptions = {}) => {
 
         var unwrappedAction = action.type === 'PERFORM_ACTION' && action.action ? action.action : action
 
+        // if matchException is true, actionType is matched with exception
+        const matchException = (exception, actionType) => {
+          if (typeof exception === 'string') {
+            return actionType === exception
+          } else if (typeof exception === 'function') {
+            return actionType === exception
```
RobCherry: This should be `return exception(actionType);`.

packages/reactotron-redux/src/create-action-tracker.js
```diff
@@ -39,8 +39,24 @@ export default (reactotron, trackerOptions = {}) => {
...
+          } else if (exception instanceof RegExp) {
+            return exception.test(actionType)
+          } else {
+            return false
+          }
+        }
+
+        const matchExceptions = exceptions.reduce((prev, exception) => matchException(exception, unwrappedAction.type) || prev, false)
```
RobCherry: Flip this so that a match will short circuit?

```
const matchExceptions = exceptions.reduce((memo, exception) => memo || matchException(exception, unwrappedAction.type), false)
```

RobCherry: Or even better, use `any` instead of `reduce`...

```
import { any } from 'ramda'
...
const matchExceptions = any(exception => matchException(exception, unwrappedAction.type), exceptions)
```

g6ling:
@RobCherry  Thanks. I applied your advice.

skellock:
Thank you!  Sorry about the slow merge.

PR #329, opened 2017-02-01
[React Native] Split `get-host` to independent module

jhen0409:
This will be used for [remotedev](https://github.com/zalmoxisus/remotedev) and [remote-redux-devtools](https://github.com/zalmoxisus/remote-redux-devtools) and here, make it as a module will be more easily to maintain React Native changes in one place.


skellock:
Wonderful!  I can't thank you enough for your help with this.

PR #332, opened 2017-02-07
Add Slackin badge

mlaco:


jamonholmgren:
Should we get rid of the reactotron-slack link? @skellock @mlaco

skellock:
Yep.  Just did.

PR #356, opened 2017-03-08
AsyncStorage plugin

rmevans9:
This is the start of an AsyncStorage plugin for React Native. This has nothing in the UI complete, just the backend implementation.

NOTE: I put just a few of the methods as callable actions in the demo react native app. The rest could be added if needed.


skellock:
That's great!  Thank you.  A nice stop gap until we get some UI in.

PR #375, opened 2017-03-15
Reactotron React Native useAllIncluded

rmevans9:
This is for all those devs out there that feel like having 5 use statments just to wire up Reactotron React Native "is just a bit over the top".

See demo for a example of using this.

This is 100% backward compatible since Reactotron React Native is still exporting each individual plugin

### DIFF
--- a/packages/demo-react-native/App/Config/ReactotronConfig.js
+++ b/packages/demo-react-native/App/Config/ReactotronConfig.js
@@ -24,6 +24,11 @@ if (__DEV__) {
       reconnectionAttempts: 10
     }
   })
+    .useAllIncluded({
+      trackGlobalErrors: { veto: frame => vetoTest(frame.fileName) },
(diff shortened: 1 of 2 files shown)


packages/demo-react-native/App/Config/ReactotronConfig.js
```diff
@@ -24,23 +24,19 @@ if (__DEV__) {
       reconnectionAttempts: 10
     }
   })
+    .useAllIncluded({
```
jaimeagudo: I just tested this and i got a red screen `undefined is not a function (near '...useAllIncluded({...')` 
First stacktrace points exactly to this line `ReactotronConfig.js:27`

skellock: run ./copy-internal-deps.sh in that directory.

PR #383, opened 2017-03-22
Fix timeline filtering for 'Subscriptions Changed'

javiercr:
As discussed in Slack the subscription changed filter is ignored (i.e. if you uncheck it, subscription changed events will still appear).

I think this should fix it but I haven't had a chance to test it yet. @skellock could you confirm it works? Otherwise tomorrow I'll try to setup the dev environment for reactotron and test it myself.


rmevans9:
@skellock gave this a quick test locally and it worked as I expected.

LGTM!

javiercr:
Thank you @rmevans9!

PR #388, opened 2017-03-31
Takes haul for a spin.

skellock:
I was able to get this up & running with [haul](https://github.com/callstack-io/haul), but there's a few issues:

1. sourcemap lookups are broken because it looks like they haven't implemented that end point yet in haul
2. openInEditor plugin is broken as well, probably same reason
3. had a strange collision with due to the `reactotron-react-native` package having a devDependency of `react-native` and not able to be located.  I ended up symlinking the `packages/demo-react-native/node_modules/react-native` over, but that's no fun.

Great news is:  once up & running, the symlink provided by lerna "just worked".  

I'm very excited to use it.  There's no reason we can't use it for some types of development already (just not errors or openInEditor).


grabbou:
That's exciting! There's definitely plan to support those missing features in the nearest future. Feel free to open issues in `haul` repo, will make it easier to track updates.

skellock:
@grabbou Ya, sure dude.  I'll open a few more issues/prs shortly.  Been crazy with client work lately and then a short vacation.

PR #396, opened 2017-04-16
assisting packages and docs

GantMan:
So I came back to try to contribute to Reactotron, and keep in mind I'm like "ER MA GERD WUT?" so please forgive me if any of this is wrong/strange.

So there were TONS of deps that were missing for me.  Every time I ran a command from the doc, it was like "BWAHAHAHA, I aint doin' shit!"  It made me feel kinda bad about life.  So I went to the psychiatrist, and she said, I needed to go through reconciliation with a moment of truth or I'd keep wetting the bed... long story longer... I decided to make this pull request, knowing I'm probably adding deps that couldn't really be missing.  But in my own strange 6th sense kind of way they really are.  Anyhow, this is how I got it commands to work on my machine.

🦄

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -16,6 +16,7 @@ npm install
 npm run welcome
 ```
 
+You may need to install `rollup` globally.
 
 ### Code Style
 
@@ -50,7 +51,7 @@ How we current get around this is by:
 
 Many tears were shed and this is the least shitty solution I could muster.
 
-In the root, you can run `npm copy-internal-deps` to make them copy over.
+In the root, you can run `npm run copy-internal-deps` to make them copy over.
 
 But since you're copying built dependencies, you'll need to run `npm run build` first
 if you're doing this to bring forward changes in those dependency libraries.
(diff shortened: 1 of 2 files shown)


docs/contributing.md
```diff
@@ -16,6 +16,7 @@ npm install
 npm run welcome
 ```
 
+You may need to install `rollup` globally.
```
skellock: You don't.  Which package was causing this?

skellock:
Let's chat in person tomorrow about what was going wrong.  None of these changes should be required to build.

GantMan:
It's outside the history of my terminal now.  Bc I'm just typing up a storm.  But it didn't work.   Installing these made it work *shrug*

docs/contributing.md
GantMan: will round it up with you later and see.

skellock:
Alright. I'll close this PR up though. Root dependencies aren't the right spot for these dependencies.

PR #413, opened 2017-04-22
[Needs Help] - Rename state

GantMan:
Live from Amsterdam, almost functioning PR needs lil help
snapshots


GantMan:
Just need to know how to mobx that data back to where it needs to go.  Maybe a quick screen sesh would do it.  I kinda see what we do, but not too sure.

skellock:
Sweet.  Now [something like this](https://github.com/infinitered/reactotron/blob/master/packages/reactotron-app/App/Stores/UiStore.js#L219) but for renaming?

GantMan:
I'm a little lost on the `this.server.commands['state.backup.response'].remove(state)` since I can't seem to figure out what commands are.  I could pass the backup around and then modify it in the final event, but that seems dirty.  I'm not sure what the best practice should be.

skellock:
The commands come from reactotron-core-server. Its a map of command names to the latest 100 messages for each.

Not sure what you're getting at about best practices. Is there better ways to build this? Absolutely.

This works pretty good. Has for a year. I'm all ears if you're thinking of refactoring.

skellock:
renamesnapshot

GantMan:
lookin good

PR #424, opened 2017-05-10
Comment out sending of the effect tree for now.

rmevans9:
For now I am just commenting out the relevant code since we will want to revisit this and might need it back.


skellock:
Thx @rmevans9 .  I think this is how we transfer our redux-saga info tho isn't it?

rmevans9:
https://github.com/infinitered/reactotron/blob/master/packages/reactotron-redux-saga/src/saga-monitor.js#L141

This is where it actually sends the saga info upon completion of the saga which is the functionality you see in Reactotron currently.

skellock:
I'm dumb and I shouldn't be allowed to type on a keyboard.

PR #461, opened 2017-06-16
Typescript Conversion - core-server && core-client

rmevans9:
This is to show what a really really basic TypeScript conversion of one of the packages took.

Things to note:
- Ramda typings can be "hard" because of the way the library works. Might run into issues with more complex usage. Note that I already found one that I had to work around
- This is not trying to build good types at all... more of just an exercise of "can we make it run and compile good code that works without any changes outside of the package in question".


packages/reactotron-core-client/package.json
```diff
@@ -24,19 +24,16 @@
...
-    "babel-eslint": "^7.2.3",
-    "babel-preset-es2015": "^6.24.1",
-    "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-preset-es2015-rollup": "^3.0.0",
-    "babel-preset-stage-1": "^6.24.1",
     "nyc": "^10.3.2",
     "rollup": "^0.41.6",
-    "rollup-plugin-babel": "^2.6.1",
-    "standard": "^10.0.2",
+    "rollup-plugin-typescript2": "^0.4.3",
```
skellock: very cool

packages/reactotron-core-client/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "noImplicitAny": false, // Might be worth setting to true after we get good typing in
```
skellock: Sure, wanna go strict mode for now?  I think it excludes this guy.

packages/reactotron-core-client/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+        "removeComments": false,
+        "preserveConstEnums": true,
+        "sourceMap": true,
+        "allowSyntheticDefaultImports": true,
```
skellock: BOOOOOOOOOoooooooOOooOOoooooo.  I actually don't know why I'm booing.

skellock:
You wanna go 2.4rc for the string enums?

packages/reactotron-core-client/tsconfig.json
rmevans9: I don't like that I had to add that but some modules were not wanting to come in.

rmevans9:
I can go RC on TypeScript, no problem.

packages/reactotron-core-client/tsconfig.json
rmevans9: I am open to that. I was just trying to go the path of least resistance here so I could test if the output was actually able to be dropped in and used.

packages/reactotron-core-client/package.json
rmevans9: There were a few around... that one seemed to be the most up to date one.

packages/reactotron-core-client/tsconfig.json
skellock: Currently experiencing this problem on my client project with a few libs.  I can `import * as ValidateJS from 'validate.js'`, but it won't work in the `jest` environment.  I think somethings borked with the typings on that library.

But I digress.

GantMan:
For anyone using Reactotron and TypeScript and tired of seeing all the typescript warnings on `console.tron...` you can fix it with just a few lines.

Make this `types/console.index.d.ts`
```
interface Console {
  tron
}
```

And add this to `tsconfig.json`
```
  // Allows you to define yoru own types in /types folder
  "baseUrl": "types",
  "typeRoots": ["types"]
```

It might be worth it to ship this definition in the next release so those using typescript can avoid errors until the conversion is complete.

rmevans9:
I can do that @GantMan. I will work on adding that next time I get a chance to work on this PR.

rmevans9:
First pass at adding some good typing to `reactotron-core-server` is done. I didn't touch the main part of the package yet because I have to work through some parts of it still - like what a command actually looks like so we can get some really good definitions going. Might be a project for after we merge this though.

PR #467, opened 2017-06-28
Make apisauce resolve to ^0.14.0 to make React Native work again.

admmasters:


rmevans9:
I don't see a problem with updating the apisauce version however can you provide a bit of detail as to why 0.13.0 was making this not work on React Native? Personally I am not running into any issues.

skellock:
I think this has to do with React Native 0.45.1 and the new warning.  https://github.com/skellock/apisauce/issues/108

PR #472, opened 2017-07-02
Change 'Hide Others' accelerator to standard macOS shortcut

andrewsardone:
I just started use Reactotron and I'm digging it! I noticed this inconsistency, so I thought I'd send over a PR.

The old Electron accelerator for the `Hide Others` command was `Command+Shift+H`, but this change makes the accelerator the conventional `Option+Command+H`. See [Apple's docs][as]:

> To view the front app but hide all other apps, press Command-Option-H.

As another example, compare [Chrome's shortcut][ch] with [Reactotron's][ro].

Thanks for the neat project!

[as]: https://support.apple.com/en-us/HT201236
[ch]: https://www.dropbox.com/s/0v3flw2m2kda7pf/chrome-hide-others.png?raw=1
[ro]: https://www.dropbox.com/s/5fwsrear8eis4k8/reactotron-hide-others.png?raw=1

### DIFF
--- a/packages/reactotron-app/main.development.js
+++ b/packages/reactotron-app/main.development.js
@@ -42,7 +42,7 @@ app.on('ready', () => {
         { label: 'About Reactotron', selector: 'orderFrontStandardAboutPanel:' },
         { type: 'separator' },
         { label: 'Hide Reactotron', accelerator: 'Command+H', selector: 'hide:' },
-        { label: 'Hide Others', accelerator: 'Command+Shift+H', selector: 'hideOtherApplications:' },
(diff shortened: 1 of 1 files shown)


rmevans9:
Seems straight forward but I would be a little concerned making this change without labeling it breaking since the old keybind won't work anymore. I am wondering if this might make more sense in the v2 branch?  Thoughts @skellock?

rmevans9:
Thanks for the PR @andrewsardone!

skellock:
Oops. I didn’t realize I’d picked the wrong combo.  Thanks for pointing this out.

It is breaking, but not in the sense that it’ll make dependencies stop working.  

Also goes to show you how much I use “hide others”.

PR #473, opened 2017-07-02
remove extra line about installation

newswim:
Looks like this got added twice!


rmevans9:
LGTM! Thanks for the contribution!

Issue #485, opened 2017-07-21
[v2] Lint from the top level of the repo fails


rmevans9:
It looks like the TypeScript changes caused a failure when running lint. It complains about not being able to find module 'babel-eslint' (which makes sense since we don't use babel any more....). Needs investigation into plan to resolve.

rmevans9:
After about 30 seconds of looking into why its because of StandardJS. It does not support TypeScript which means we have to move to TSLint or similar. @skellock opinions on a TSLinter to integrate?

skellock:
I’m using this on a client project. So far so good.

```json
{
  "extends": ["tslint-react-recommended"],
  "rules": {
    "ordered-imports": [false],
    "quotemark": [true, "single", "jsx-single", "avoid-escape"],
    "semicolon": [true, "never"],
    "member-access": [false],
    "member-ordering": [false],
    "trailing-comma": [true, { "singleline": "never" }]
  }
}
```

And in dev deps:

```json
    "tslint": "5.5.0",
    "tslint-eslint-rules": "4.1.1",
    "tslint-react": "3.0.0",
    "tslint-react-recommended": "1.0.9",
```

I like it. I’m open to anything though. The only things I really care about is that it works well with`prettier` and doesn’t have the devils punctuation: the semi-colon.

rmevans9:
![screen shot 2017-07-28 at 9 44 07 pm](https://user-images.githubusercontent.com/14151327/28741128-e8689b56-73dd-11e7-9031-9ca4583e06e9.png)

I have integrated based upon the rules you suggest and there are some stylistic changes it wants like no space before a functions ( and such. Are you cool with these?

Working on `prettier` now

skellock:
Yes sir.

skellock:
Linting is working now.  I added a prettier file to the root. I’ll do a sweep with prettier in the morning.

PR #490, opened 2017-07-29
[TypeScript] Linting fixes

rmevans9:
This sets up linting for the TypeScript packages. Nothing with `prettier` yet as I am not sure if we want to integrate that into the commit flow (precommit hooks) or just rely on the developer to use prettier.

### DIFF
--- a/packages/reactotron-core-server/src/index.ts
+++ b/packages/reactotron-core-server/src/index.ts
@@ -36,32 +36,32 @@ class Server {
   /**
    * Holds the commands the client has sent.
    */
-  commands = new Commands();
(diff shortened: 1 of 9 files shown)


rmevans9:
Fixes #485

packages/reactotron-core-server/src/index.ts
```diff
@@ -36,32 +36,32 @@ class Server {
   /**
    * Holds the commands the client has sent.
    */
-  commands = new Commands();
+  commands = new Commands()
```
skellock: So long sucka!

PR #493, opened 2017-08-02
add guard for RN 0.47

GantMan:
Breaking change causes reactotron to redscreen.  This fixes.


GantMan:
Side note:  it sucks to have transpiled code ignored on github.  I can't have a branch be used in the meantime. :(

admmasters:
:) I was just about to look at submitting this

rmevans9:
@GantMan If you update the `main` pointer in the package.config of a branch you should be able to use everything with out transpiling since the react-native packager will transpile node_modules as well.

This is going to break benchmarks but to get Reactotron working again I think that is fine. We will just need to revisit and resolve how to do benchmarks for 47 sometime.

GantMan:
@rmevans9 - I made a separate ticket to track the benchmark break.

skellock:
oh crap!  thanks!

PR #506, opened 2017-09-03
Update quick start docs for React Native

evturn:


docs/quick-start-react-native.md
```diff
@@ -19,12 +19,24 @@ npm i --save-dev reactotron-react-native
...
+
+  Reactotron
+    .configure({
+      name: 'React Native Demo'
+    })
+  .use(trackGlobalErrors())
+  .use(openInEditor())
+  .use(overlay())
+  .use(asyncStorage())
+  .use(networking())
```
rmevans9: Why are you suggesting to do them individually vs the `.useReactNative()` way? It might be worth calling out that you can do them individually but I think `.useReactNative()` makes it easier.

evturn: I 100% agree, but in my experience it isn't uncommon that `.useReactNative()` throws saying that it is not a function. Though when using these individually everything checks out. After reading through some of the repo issues it appears this solution has been suggested to help users avoid this error. For what it's worth, the most recent project I've used this with is using v1.12.2.

rmevans9:
Since this was completed in another PR I am closing this one.

PR #528, opened 2017-10-10
Fixes tests and only runs projects with em.

skellock:
After rerunning setup apisauce's axios header changed.  This is fragile, so I removed that line.

The other thing here is, two of our libs don't have tests, so at the top `package.json`, we're not trying to run those anymore.


skellock:
Also, i ran `yarn lerna bootstrap`, but `package.json.lock` got changed.... lol.

PR #529, opened 2017-10-10
Switches to yarn and updates instructions.

skellock:
Ok, this straightens out the package manager drama.

Two issues being fixed here:

* `lerna` now uses `yarn`.

* `--ignore-engines` is a stop gap because one of our older dependencies in `reactotron-app` (`jss`) is using a `>= 4.x` to `<=6.x` range in their `engines` in `package.json`.  :(   This will go away when we upgrade the modal library.  Which we can't right now because of their breaking changes... lol.

For the first time in a long time... i can just type `yarn welcome` and it works first try.

Oi.


skellock:
And now it's failing with the checksum again.  🎉

PR #537, opened 2017-10-14
Upgrade Webpack + fix core-client and core-server builds

rmevans9:
I had to fix some issues with core-client and core-server to get it to build correctly. We should probably look at those two packages again as it seems they have some tslint issues.


rmevans9:
This is all the upgrades I could do safely without things breaking terribly.

PR #538, opened 2017-10-14
Add TypeScript and convert a file

rmevans9:
NOTE: This is pointed at the webpack-update branch so it gains the dep upgrades. This should only be merged after webpack-update is.


packages/reactotron-app/App/index.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react'
+import { render } from 'react-dom'
+import * as App from './Foundation/App'
+import './app.global.css'
+import * as injectTapEventPlugin from 'react-tap-event-plugin'
+injectTapEventPlugin()
+
+const ShutUpTypeScript: any = App;
```
rmevans9: Without putting app in this var and telling TypeScript it was an any the render was complaining about App not having a constructor or some such nonesense. I suspect it is related to the fact that App still JS.

skellock: ROFL

PR #542, opened 2017-10-14
No mo mobx in core

rmevans9:
This has only been lightly tested. Commands are piping into Reactotron correctly still. Minimal change footprint to get it "working". We should discuss further changes to this PR


rmevans9:
For bonus points this also adds the emitter and uses that instead and also purges out some methods that could live up in reactotron app. The intent with that change is most of these commands will be heading to live in plugins so should not be in core at all.

skellock:
I like it.  Good first step towards plugins.  Thanks man.

PR #549, opened 2017-10-31
Made the instructions more clear.

serranoarevalo:
Tried to integrate it with redux and it was a little bit confusing to use Reactotron.createStore, I wasn't sure if I had to import Reactotron from the npm package or the one I'm creating on the file ReactotronConfig.js

### DIFF
--- a/docs/plugin-redux.md
+++ b/docs/plugin-redux.md
@@ -46,6 +46,8 @@ Reactotron
   .configure({ name: 'React Native Demo' })
   .use(reactotronRedux()) //  <- here i am!
   .connect() //Don't forget about me!
+
+export default Reactotron;
 ```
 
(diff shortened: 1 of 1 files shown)


rmevans9:
When you `use` a plugin it attaches to the Reactotron object which comes in via the normal import. So although your suggested update would work it is not the only way to do it. You could simply `import Reactotron from 'reactotron-react-(native/js)` and do `Reactotron.createStore` off of that.

skellock:
I do agree that it is uncommon for libraries to work like this.  Reactotron acts like a self-managed Singleton. It handles its own state once configured.

Like @rmevans9 says, plugins extend the default Reactotron object once they attach via `use`.

I do very much appreciate the PR. You do make it clear that there is a problem in how we explain this though. 😊

Issue #268, opened 2016-11-17
Full text filtering


reicheltp:
It would be awesome to filter / search the logs shown in the reactotron cli.

I have something in my mind like: pressing `L` for log and enter a regex and press enter to apply it. 

If this is a welcome feature, I will start preparing a PR.

skellock:
Heyo.

I'm considering dropping support (#229) for the CLI.  There's a lot more we can do with desktop app.  If we can keep the keystroke friendly maybe we can convince more people to use it.

Having said that, I'd absolutely welcome that PR, but I'd hate for you to invest some time into something that might be going away.

I know some folks are diehard CLI'ers.  Do you consider yourself one?  Would the desktop app be a non-starter for you?  

What if we put themes in place to make you feel at home?

![image](https://cloud.githubusercontent.com/assets/68273/20396362/f8b782e8-acb3-11e6-8e50-d0acba66d914.png)

😍

reicheltp:
I am absolutely fan of GUI application. Unfortunately, I am working on Windows 😜. I have tried to compile the gui application but got some really strange errors with babel and rollup, so the cli was the easiest way to get reactotron run.

If you can help me get over this, I am the first one switching to GUI and help to improve it! 😄 

---
<details> 
  <summary>The babel rollup error:</summary>
```
Error transforming C:\tools\reactotron\packages\reactotron-core-server\src\index.js with 'babel' plugin: It looks like your Babel configuration specifies a module transformer. Please disable it. If you're using the "es2015" preset, consider using "es2015-rollup" instead.
 See https://github.com/rollup/rollup-plugin-babel#configuring-babel for more information
Error: Error transforming C:\tools\reactotron\packages\reactotron-core-server\src\index.js with 'babel' plugin: It looks like your Babel configuration specifies a module transformer. Please disable it. If you're using the "es2015" preset, consider using "es2015-rollup" i
nstead. See https://github.com/rollup/rollup-plugin-babel#configuring-babel for more information
    at preflightCheck (C:\tools\reactotron\packages\reactotron-core-server\node_modules\rollup-plugin-babel\dist\rollup-plugin-babel.cjs.js:43:102)
    at Object.transform$1 [as transform] (C:\tools\reactotron\packages\reactotron-core-server\node_modules\rollup-plugin-babel\dist\rollup-plugin-babel.cjs.js:104:18)
    at C:\tools\reactotron\packages\reactotron-core-server\node_modules\rollup\src\utils\transform.js:19:35
[...]

PR #550, opened 2017-11-02
Added a basic query field to the timeline header

Rhineheart:
Issue: #268

It will compare given query to stored command type, triggerType, name, and preview. It may display it by match or as a timeline. Timeline is default. It will show the last 100 entries in the timeline, and the last 10 per match.

_Timeline example_
screen shot 2017-11-02 at 10 42 38

_Match example_
screen shot 2017-11-02 at 10 34 07

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,11 +4,11 @@
   "version": "0.0.2",
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "clean": "lerna clean --yes",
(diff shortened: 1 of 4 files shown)


Rhineheart:
Ran the linter and removed the paths. I think I added them before even trying any commands from habit. Didn't know yarn added `./node_modules/.bin` to its environment, that's good to know!

skellock:
Lovely!  Thank you kindly!  I'll see if i can kick out another release on Sunday.

Issue #546, opened 2017-10-20
Ignore specific API Responses


Xyzor:
Hello,

I'd like to know if it's possible to ignore all the symbolicate and logs requests?

rmevans9:
This should now be possible because of #562. If you think that doesn't cover this issue please reopen.

PR #562, opened 2017-12-05
react-native networking option to skip requests based on URL.

sidnair:
I ran into an issue where logging-related API requests from tooling generated noise in Reactotron. This was reported by someone else in https://github.com/infinitered/reactotron/issues/546, but since that issue didn't generate discussion, I thought I'd send a quick PR to see if an approach like this was reasonable.

No worries if it's not!

### DIFF
--- a/docs/plugin-networking.md
+++ b/docs/plugin-networking.md
@@ -23,10 +23,13 @@ You're done.
 
 ## Advanced Usage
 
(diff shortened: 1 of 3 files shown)


skellock:
This looks great! Typings and docs. ❤️

Could we call it `ignoreUrls`?

Great feature to have. Those symbolicate calls are annoying to see haha.

sidnair:
Thanks!

FYI, I think https://github.com/infinitered/reactotron/issues/546 and https://github.com/infinitered/reactotron/issues/268 (unrelated to this diff) could be closed.

PR #568, opened 2017-12-15
Update deps

megazoll:
`react-modal-dialog` required node < 6.0.0


skellock:
I had been dodging this one because it was a breaking change.

skellock:
We have a fix for this on the `v2` branch.  Thank you for the pr!

PR #572, opened 2017-12-19
Persist timeline filter settings

megazoll:


henrikra:
I would love to have this feature!! Can you also make feature where Reactotron clears time line on reload? :)

megazoll:
@henrikra https://github.com/infinitered/reactotron/blob/master/docs/tips.md#clearing-reactotron

henrikra:
Good tip! I still think that this should be configured in Reactotron UI with checkbox for example. It does not make sense for me to change my code when I want to change settings of a running app :P

skellock:
@henrikra I agree with that.  Feel free to open another issue to track that!

henrikra:
"Woah, I had no idea that localStorage worked in electron." Well Electron is just a browser :D

skellock:
Haha... i have a long history of not putting 2 and 2 together.

PR #573, opened 2017-12-21
🔧 More Fixes the Typescript types for reactotron plugins

movibe:
Support for all reactotron plugins


skellock:
Thanks for taking the time to make this PR.  Sincerely appreciate it!

The problem I have here is that it now puts a dependency of `redux` into `reactotron-react-native` where none exists now.

I'm not sure of the right way to handle this, so suggestions are welcomed!  I haven't quite find the right way to do this.

PR #592, opened 2018-01-13
Add storybook switcher

rmevans9:
This fixes some lint issues
This fixes the custom command dialog
This introduces a storybook HOC that allows you to switch between Storybook and the main app


rmevans9:
NOTE: This doesn't add anything to the app to switch between storybook and the main app so you have to send a custom command. The commands you can send are:

`storybook:true`
`storybook:false`

PR #606, opened 2018-02-24
Connection filtering

rmevans9:
Here is a quick first pass to show that it is possible. From here I plan to do a few things:

- [x] Cleanup the UI (it looks like hell right now... it is also only on the timeline)
- [x] Contain a `clear` call from a client to only messages from THAT client. With the current changes it clears out all clients messages if any one client calls clear
- [x] Figure out how to handle subscriptions
- [ ] ~Figure out how to handle state backups~ We (aka I) just don't care about them at this point. Right now if you have multiple connections this part of Reactotron acts up anyway. Opening an issue to resolve later.


packages/reactotron-app/App/Foundation/TimelineHeader.js
```diff
@@ -101,8 +101,15 @@ class TimelineHeader extends Component {
     }
   }
 
+  handleSelectedConnectionChange = e => {
+    const connectionId = e.target.value
+    const selectedConnection = this.props.session.connections.find(c => c.id == connectionId) || null
```
rmevans9: Destructure all the things!

packages/reactotron-app/App/Foundation/TimelineHeader.js
```diff
@@ -121,6 +130,22 @@ class TimelineHeader extends Component {
           </div>
           <div style={Styles.center}>
             <div style={Styles.title}>{TITLE}</div>
+            <div>
+              <select
+                value={selectedConnectionId}
+                onChange={this.handleSelectedConnectionChange}
```
rmevans9: Style this yo.... but before that consider if this is where we actually want it to be.

packages/reactotron-app/App/Stores/SessionStore.js
```diff
@@ -152,10 +169,21 @@ class Session {
     }
   }
 
+  handleConnectionsChange = () => {
+    // TODO: See if we can be better at clearing instead of clearing everything then resetting every single time.
+    this.connections.clear()
+    this.connections.push(...this.server.connections)
```
rmevans9: I feel dirty. We are clearing and recreating the list of connections ever. single. time.

The connection in question should be getting passed in with each event so it should be easy to pluck it on out or push just it on so we are not recreating the list... however this worked for now so to just get the idea out I went with it. ¯\_(ツ)_/¯

packages/reactotron-core-server/src/reactotron-core-server.ts
```diff
@@ -98,9 +103,11 @@ export default class Server {
 
     // register events
     this.wss.on("connection", (socket, request) => {
+      const thisConnectionId = this.connectionId++
+
       // a wild client appears
       const partialConnection = {
-        id: (socket as any).id, // issue
```
rmevans9: @skellock I suspect `socketio` gave us an id but `ws` does not. So we make a incrementing one similar to message ids.

packages/reactotron-core-server/src/reactotron-core-server.ts
```diff
@@ -112,17 +119,20 @@ export default class Server {
       this.emitter.emit("connect", partialConnection)
 
       // when this client disconnects
-      socket.on("disconnect", () => {
+      socket.on("close", () => {
```
rmevans9: `ws` doesn't have a "disconnect" event. Just `close`. So lets do that instead.

packages/reactotron-core-server/src/reactotron-core-server.ts
```diff
@@ -98,9 +103,11 @@ export default class Server {
 
     // register events
     this.wss.on("connection", (socket, request) => {
+      const thisConnectionId = this.connectionId++
```
rmevans9: #namingthingsishard

packages/reactotron-core-server/src/reactotron-core-server.ts
```diff
@@ -112,17 +119,20 @@ export default class Server {
...
-          propEq("id", (socket as any).id),
+          propEq("id", thisConnectionId),
           this.partialConnections,
         ) as any
 
         // remove them from the main connections list
-        const severingConnection = find(propEq("id", (socket as any).id), this.connections)
+        const severingConnection = find(propEq("id", thisConnectionId), this.connections)
         if (severingConnection) {
-          (this.connections as any).remove(severingConnection)
```
rmevans9: Not sure when `.remove` was there but its not now so had to do it another way.

packages/reactotron-core-server/src/reactotron-core-server.ts
```diff
@@ -138,32 +148,34 @@ export default class Server {
...
           // bestow the payload onto the connection
           const connection = merge(payload, {
-            id: (socket as any).id,
+            id: thisConnectionId,
             address: partConn.address,
           })
 
           // then trigger the connection
           this.connections.push(connection)
+          this.emitter.emit("connectionEstablished", connection)
```
rmevans9: For this exercise I pull from `connections` to populate the list of connections... so I needed a new event when a connection is fully established and the intro has been made. Maybe there is a better way ¯\_(ツ)_/¯

packages/reactotron-app/App/Foundation/ConnectionSelector.js
```diff
@@ -0,0 +1,36 @@
...
+
+  render() {
+    const { connections, selectedConnection } = this.props.session
+
+    if (connections.length < 2) return null
+
+    const selectedConnectionId = selectedConnection ? selectedConnection.id : -1
+
+    return (
+      <select value={selectedConnectionId} onChange={this.handleSelectedConnectionChange}>
```
rmevans9: This needs styled but ¯\_(ツ)_/¯ how. Maybe leave it until @skellock gets fed up with the out of place `<select>` and styles it for me?

packages/reactotron-app/App/Foundation/ConnectionSelector.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from 'react'
+import { inject, observer } from 'mobx-react'
+
+@inject('session')
```
rmevans9: This is a component of sorts that connects to `mobx`. It feels a little bit dirty but it buys us the ability to just dump this component on screen and pretend everything is fine (mainly because everything is fine...)

packages/reactotron-app/App/Foundation/SubscriptionsHeader.js
```diff
@@ -71,6 +72,9 @@ class SubscriptionsHeader extends Component {
           </div>
           <div style={Styles.center}>
             <div style={Styles.title}>{TITLE}</div>
+            <div>
+              <ConnectionSelector />
```
rmevans9: TODO: Build out filtering out the state based on connection.
TODO: Figure out the best way to show state trees when "ALL" is selected... is that even possible. WOULD YOU ACTUALLY WANT THAT YOU CRAZY?

packages/reactotron-app/App/Stores/SessionStore.js
```diff
@@ -144,18 +155,45 @@ class Session {
...
 
+  // Sets the selected connection
+  @action
+  setSelectedConnection (connection) {
+    this.selectedConnection = connection
+  }
+
   handleCommand = command => {
     if (command.type === 'clear') {
+      const newCommands = pipe(
```
rmevans9: @skellock This might suck... yeah... I think this sucks. I don't think `pipe` is needed to do a single thing like this but when calling reject directly it threw a fit. If you have thoughts on a better way I am allllll ears.

rmevans9:
@skellock RE State backups - how do you think we should handle this? Right now it should (haven't tested) backup and restore to all connected devices... I am thinking that might be ok for now. Thoughts?

packages/reactotron-core-server/src/reactotron-core-server.ts
skellock: oops

packages/reactotron-app/App/Foundation/ConnectionSelector.js
skellock: That's fine.

packages/reactotron-app/App/Foundation/ConnectionSelector.js
skellock: So far so good.  Sign me up for the UI parts.

packages/reactotron-app/App/Foundation/SubscriptionsHeader.js
skellock: I think we pick 1 "read/write", and the others are "write-only".

packages/reactotron-app/App/Stores/SessionStore.js
skellock: That's weird.  Feel free to convert anything basic like this to non-ramda.  `.filter`, `.map`, `.sort`... i think these are all preferable to use the built-in functions rather than to shoe-horn ramda in there.

packages/reactotron-app/App/Stores/SessionStore.js
skellock: That's fine.  We're in dev mode.  We're not optimizing for performance.

packages/reactotron-core-server/src/reactotron-core-server.ts
skellock: Nailed it.

packages/reactotron-core-server/src/reactotron-core-server.ts
skellock: Oops.

packages/reactotron-core-server/src/reactotron-core-server.ts
skellock: Oops.

packages/reactotron-core-server/src/reactotron-core-server.ts
skellock: Oops.

rmevans9:
I am BACK... almost exactly two months later to wrap this baby up. I updated the UI to fit better with the Reactotron style. I think this is good to go. Note the updated original comment. Specifically around the state backups - they will take more work then I think I want to shove into this PR as it is already pretty full. I will open an issue and very likely fix that issue this week.

rmevans9:
WIP tag removed. LETS GOOOO!

rmevans9:
screen shot 2018-04-26 at 8 40 51 pm
screen shot 2018-04-26 at 8 41 00 pm

rmevans9:
Closing in favor of a new one that will come with a better UI that fits in with the direction this is headed.

PR #616, opened 2018-03-12
✨ Adds milliseconds to the timeline

### LINKED ISSUE #614: Feature Request: Timeline should always show milliseconds

skellock:
## Overview

The time isn't as granular as it should be on the timeline.  It's meaningful to see if something took 1ms or 99ms.

This came up in #614 but I coincidentally experienced this last week too.

## Screenshots

image

## Other Notes

Moving from the heavyweight `moment` to lightweight `date-fns`.  Calling `moment()` incurs some significant overhead surprisingly!

I also made the `h:mm:` part a little darker than it was before.

Fixes #614

### DIFF
--- a/packages/reactotron-app/App/Shared/Timestamp.js
+++ b/packages/reactotron-app/App/Shared/Timestamp.js
@@ -1,26 +1,19 @@
-import React from 'react'
-import PropTypes from 'prop-types'
-import moment from 'moment'
-import Colors from '../Theme/Colors'
+import React from "react"
+import PropTypes from "prop-types"
+import moment from "moment"
+import Colors from "../Theme/Colors"
+import { format } from "date-fns"
 
 const Styles = {
-  container: {
-    margin: 0,
-    padding: 0
-  },
-  left: {
-    color: Colors.foregroundDark
-  },
-  right: {
-    color: Colors.foreground
-  }
+  container: { margin: 0, padding: 0 },
+  left: { color: Colors.highlight },
+  right: { color: Colors.foreground },
 }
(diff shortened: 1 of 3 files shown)


petermjones:
Thanks for the change.  

"It's meaningful to see if something took 1ms or 99ms"

You mean if something took 10ms to 990ms, since we are only showing 2 decimal places of a 3 decimal place value?  It would be perfect if you could show all 3 places.

skellock:
Sub-milliseconds times aren’t meaningful in this lib since we have to deserialize the messages.  You might be interested in `benchmark` when you get down that low though.

petermjones:
Is the time displayed when the message is received?  Maybe it should be part of the serialized message for improved precision?  4 digits would be sub-millisecond precision, 3 digits would be milliseconds.  In any case, an improvement over what it was yesterday!

skellock:
Oh man, I totally missed that. You’re right. I’ll do a pr in the morning.  🤣

skellock:
The time is assigned server side, [post deserialize](https://github.com/infinitered/reactotron/blob/master/packages/reactotron-core-server/src/reactotron-core-server.ts#L136). Client side would be more acurate, agreed.

petermjones:
To me, having client-side timing would be a big plus when it comes to optimizing an app so it behaves like it was fully native.  Hard to optimize what you cannot see or accurately measure.  ReactoTron, for backwards compatibility, could just detect if the time was passed, if it's not use the internal/local time.

skellock:
#621 has these changes

PR #621, opened 2018-03-13
✨ Adds more precision and shows delta times

skellock:
## Overview

Improves timestamp information.

## Screenshot

image

## Details

3 changes here:

1. Display 3 digits (up from 2) of milliseconds in the time.
2. Display a smaller time delta in timeline. Time since last Reactotron message.  Since it's relative, I made it small because it won't always matter.
3. Moving timestamps **from the server to the client** for better accuracy.  We can now see small changes like 1ms and 2ms.  I'm shocked this made such a big difference! And shocked again that this is shocking news to me.

Thanks for the tips @petermjones!

### DIFF
--- a/packages/reactotron-app/App/Commands/ApiResponseCommand.js
+++ b/packages/reactotron-app/App/Commands/ApiResponseCommand.js
@@ -143,7 +143,7 @@ class ApiResponseCommand extends Component {
     const summary = { 'Status Code': status, Method: method, 'Duration (ms)': duration }
 
     return (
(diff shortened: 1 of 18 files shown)


skellock:
Slight change in plans.

What you see in the timeline is a subset of the messages `Reactotron` sends.  In addition, it's possible to filter out certain types of messages.

If the math doesn't compute from row-to-row, people (me!) are going to get confused.

So, that screenshot you see above is "technically" correct, it doesn't make sense.

This one makes sense though:

image

petermjones:
Looks good, thanks.

PR #637, opened 2018-04-25
Adds better image support

skellock:
@GantMan pointed out some problems with the displaying of images.

The typings for `display` was wrong.  it said `image` was a string.  it was actually an object.

It makes sense to want to just show a string, so this adds supports for both.

I also added the missing typings for `image()`.


skellock:
He's at a conference, so I'm fast tracking this thru.

skellock:
Nothing could possible go wrong.  Nope.

skellock:
This is fine. 

🔥🔥 🐶 ☕️  🔥🔥

rmevans9:
https://media.giphy.com/media/z9AUvhAEiXOqA/giphy.gif

PR #641, opened 2018-04-27
Better custom commands RFC example + a few other minor fixes

rmevans9:
Pointed at connection filtering because that was my base branch when I decided to see what I could come up with for this... so these changes should be merged after that PR is merged.

Things I did:
- [x] Setup basic custom command registration/unregistration
- [x] Setup calling registered commands
- [x] Ship commands over the wire when commands are register/unregistered
- [x] Update an observable with the list of available commands
- [x] Fix random warnings in the demo app
- [x] Test that registered custom commands work with the current custom command dialog

Things I didn't do:
- [ ] Setup the Reactotron UI to show the available commands
- [ ] Correctly name the commit message (rfp instead of rfc? What was I thinking?)
- [ ] Make all the code clean and conform to standards (see TODOs in code)

### DIFF
--- a/packages/demo-react-native/App/Config/ReactotronConfig.js
+++ b/packages/demo-react-native/App/Config/ReactotronConfig.js
@@ -28,4 +28,7 @@ if (__DEV__) {
   Reactotron.clear()
 }
 
+const testUnreg = Reactotron.onCustomCommand('test', () => { console.log('I WORK!') })
+Reactotron.onCustomCommand('test2', () => { testUnreg() })
+
 console.tron = Reactotron
--- a/packages/reactotron-app/App/Stores/SessionStore.js
+++ b/packages/reactotron-app/App/Stores/SessionStore.js
@@ -46,6 +46,7 @@ class Session {
   @observable connections = []
   // Selected Connection
   @observable selectedConnection = null
+  @observable customCommands = []
 
   commandsManager = new Commands()
 
@@ -179,6 +180,15 @@ class Session {
 
       this.commandsManager.all.clear()
       this.commandsManager.all.push(...newCommands)
+    } else if (command.type === 'customCommand.register') {
+      this.customCommands.push(command.payload)
+    } else if (command.type === 'customCommand.unregister') {
+      const newCustomCommands = pipe(
+        dotPath('customCommands'),
+        reject(c => c.id === command.payload.id),
+      )(this)
+      this.customCommands.clear()
+      this.customCommands.push(...newCustomCommands)
     } else {
       this.commandsManager.addCommand(command)
     }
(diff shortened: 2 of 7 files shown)


packages/demo-react-native/App/Config/ReactotronConfig.js
```diff
@@ -28,4 +28,7 @@ if (__DEV__) {
   Reactotron.clear()
 }
 
+const testUnreg = Reactotron.onCustomCommand('test', () => { console.log('I WORK!') })
+Reactotron.onCustomCommand('test2', () => { testUnreg() })
```
rmevans9: If this is the direction we go I would think we would want to setup a better demo but this serves its purpose - which is to simply show off that I can write code that works all of then time...60% of the time.

packages/reactotron-app/App/Stores/SessionStore.js
```diff
@@ -179,6 +180,15 @@ class Session {
 
       this.commandsManager.all.clear()
       this.commandsManager.all.push(...newCommands)
+    } else if (command.type === 'customCommand.register') {
```
rmevans9: We might want to declutter this method a little bit as we get more and more "built in" commands. Maybe have a `BuiltInCommandHandler` of some sort that you pass the command and it just returns `true/false` if we should add the command on to the commandsManager

packages/reactotron-core-client/src/reactotron-core-client.ts
```diff
@@ -291,6 +313,30 @@ export class Client {
...
+      id: this.customCommandLatestId,
+      command,
+      handler,
+    }
+
+    // Increment our id counter
+    this.customCommandLatestId += 1
+
+    // Add it to our array
+    this.customCommands.push(customHandler)
```
rmevans9: We track custom commands on client AND server. On client we have all the metadata + the handler but on server we only have metadata. The way I just shovel over each registration/unregister over to the server individually could lead to a desync between the client and server. Might want to consider just shipping all metadata of all registered commands every single time.... but then that bloats the payload. I don't know what way is the lesser of two evils. Why must keeping things in sync be so hard?

skellock:
Checking this out out @rmevans9 ! ❤️

packages/demo-react-native/App/Config/ReactotronConfig.js
skellock: haha

packages/reactotron-core-client/src/reactotron-core-client.ts
```diff
@@ -161,6 +177,12 @@ export class Client {
 
       // trigger our plugins onCommand
       forEach(plugin => plugin.onCommand && plugin.onCommand(command), this.plugins)
+
+      // trigger out register custom commands
+      if (command.type === "custom") {
+        // TODO: Ramda-tize
```
skellock: I prefer this way.  I've fallen of the `ramda` train for straight-forward statements like this.  💯

rmevans9: :+1:

rmevans9:
@skellock please feel free to hop on this branch and tweak as you see fit. It is based off an idea you put out there after all!

PR #643, opened 2018-04-27
Storybook now has UI in Reactotron!

rmevans9:
This adds a really basic UI for the storybook toggle feature. Its dirty. It should be cleaned up which I will do on another day. Just getting this out there to be looked at/judged by others.

**Is this a breaking change?** eh... is it really breaking if you never told anyone the feature was there in the first place?

**What does it look like?** 
screen shot 2018-04-26 at 10 07 16 pm

**Is the active state 100% in line with the actual state of if storybook is shown in the app?!?** No. State sync between client and server is not always the most straight forward thing. Could we make this better? Yes... but we should do it in a more efficient manner then just solving it in one offs for everything we are starting to need to keep in sync between client and server I think.

### DIFF
--- a/packages/reactotron-app/App/Foundation/Native.js
+++ b/packages/reactotron-app/App/Foundation/Native.js
@@ -65,6 +65,20 @@ const Styles = {
   overlayPreview: {
     maxWidth: '100%',
     maxHeight: '100%'
+  },
+  button: {
+    height: 30,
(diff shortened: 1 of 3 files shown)


rmevans9:
@GantMan tagging you since you voiced interest in this feature while I was out on vaca. This should make the use of this feature better.

PR #650, opened 2018-05-03
[WIP] Ramda be gone

rmevans9:
This is the start of clearing ramda out of the client side for Reactotron. There is still things to do:

- [ ] Actually test these changes to make sure they work as expected. As of writing this is 100% untested. I made sure it built. That is all. I will be testing this.
- [ ] All of the plugins
- [ ] Think deeply about `reactotron-react-js` and how much we really care about it. Then act upon that level of care.

### DIFF
--- a/packages/reactotron-apisauce/package.json
+++ b/packages/reactotron-apisauce/package.json
@@ -5,11 +5,10 @@
   "main": "dist/index.js",
   "scripts": {
     "test": "ava",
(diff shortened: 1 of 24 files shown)


rmevans9:
A quick test and nothing blew up all over me... so there is that :+1: Further testing to come (also if you feel like you want to test just let me know how it goes)

rmevans9:
👏 I forgot about Rollup configuration...

PR #653, opened 2018-05-06
[WIP] Errors Per a Second

rmevans9:
This is pointed at the connection filtering branch because I rely on an event added in the PR to track when connections are started/finished.

This is a really really basic implementation of Errors Per a Second (EPS). The UI is just thrown in the header to test the functionality of tracking this information. As of right now the time connected only updates when a connection is lost but errors per a second are updated at the time of seeing an error. This can lead to slightly off EPS scores but I think that is a solvable problem.

There are a few things I think we should do still:
 - [ ] Add this to its own tab with a reset button and all
 - [x] Figure out making the connection time count up in real time so that the EPS score is also in real time
 - [x] Handle saving all the data if Reactotron is shutdown. Right now if a connection is open and you close Reactotron you lose the current sessions time so your EPS score would be skewed.
 - [ ] Clean up the code. This is 100% hacked together (It is 11:13pm and I started writing this code at 10:56ish sooo...yeah)
 - [x] Add Errors Per Minute (EPM)
 - [x] Add Errors Per Hour (EPH)
 - [ ] Other things? ¯\\\_(ツ)\_/¯

### DIFF
--- a/packages/reactotron-app/App/Foundation/TimelineHeader.js
+++ b/packages/reactotron-app/App/Foundation/TimelineHeader.js
@@ -124,9 +124,11 @@ class TimelineHeader extends Component {
             <div style={Styles.title}>{TITLE}</div>
             <div>
               {/* <ConnectionSelector /> */}
-              conTime: {devTracker.connectionTime}
-              currErr: {devTracker.currentErrors}
+              conTime: {devTracker.connectedSeconds}
+              currErr: {devTracker.totalErrors}
               EPS: {devTracker.errorsPerSecond}
+              EPM: {devTracker.errorsPerMinute}
+              EPH: {devTracker.errorsPerHour}
             </div>
           </div>
           <div style={Styles.right}>
--- a/packages/reactotron-app/App/Foundation/TimelineHeader.js
+++ b/packages/reactotron-app/App/Foundation/TimelineHeader.js
@@ -103,7 +103,7 @@ class TimelineHeader extends Component {
   }
 
   render () {
-    const { ui, devTracker } = this.props.session
+    const { ui } = this.props.session
     const { isTimelineSearchVisible } = ui
     const searchIconStyle = {
       ...Styles.searchIcon,
@@ -123,12 +123,7 @@ class TimelineHeader extends Component {
           <div style={Styles.center}>
             <div style={Styles.title}>{TITLE}</div>
             <div>
-              {/* <ConnectionSelector /> */}
-              conTime: {devTracker.connectedSeconds}
-              currErr: {devTracker.totalErrors}
-              EPS: {devTracker.errorsPerSecond}
-              EPM: {devTracker.errorsPerMinute}
-              EPH: {devTracker.errorsPerHour}
+              <ConnectionSelector />
             </div>
           </div>
           <div style={Styles.right}>
--- a/packages/reactotron-app/App/Foundation/TimelineHeader.js
+++ b/packages/reactotron-app/App/Foundation/TimelineHeader.js
@@ -103,7 +103,7 @@ class TimelineHeader extends Component {
   }
 
   render () {
-    const { ui } = this.props.session
+    const { ui, devTracker } = this.props.session
     const { isTimelineSearchVisible } = ui
     const searchIconStyle = {
       ...Styles.searchIcon,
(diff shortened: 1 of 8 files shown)


rmevans9:
Probably worth noting that this only works with React Native right now.

packages/reactotron-app/App/Foundation/TimelineHeader.js
```diff
@@ -123,7 +123,10 @@ class TimelineHeader extends Component {
           <div style={Styles.center}>
             <div style={Styles.title}>{TITLE}</div>
             <div>
-              <ConnectionSelector />
+              {/* <ConnectionSelector /> */}
+              conTime: {devTracker.connectionTime}
+              currErr: {devTracker.currentErrors}
+              EPS: {devTracker.errorsPerSecond}
```
rmevans9: Temporary UI FTW!

packages/reactotron-app/App/Stores/DevTrackerStore.js
```diff
@@ -0,0 +1,77 @@
+import { computed, observable, action, asMap } from "mobx"
```
rmevans9: I think this file will be going through a lot of iteration and cleanup.

packages/reactotron-app/App/Stores/DevTrackerStore.js
```diff
@@ -0,0 +1,77 @@
...
+
+    constructor() {
+        const currentDevTime = localStorage.getItem(TIME_NAME)
+        const currentErrors = localStorage.getItem(ERROR_NAME)
+
+        if (typeof currentDevTime === 'undefined' || currentDevTime === null) {
+            localStorage.setItem(TIME_NAME, 0)
+        } else {
+            this.storedTime = parseInt(currentDevTime, 10)
+        }
```
rmevans9: If we have a value in local storage use it, if not lets set it to 0 just so we have SOMETHING.

packages/reactotron-app/App/Stores/DevTrackerStore.js
```diff
@@ -0,0 +1,77 @@
...
+    @computed
+    get connectionTime() {
+        let currentSessionTime = 0
+
+        if (this.sessionStarted) {
+            const currentTime = new Date()
+            currentSessionTime = (currentTime.getTime() - this.sessionStarted.getTime()) / 1000
+        }
+
+        return this.storedTime + Math.floor(currentSessionTime)
```
rmevans9: This exists because we have the time of all previous sessions but the time of the current session is not in there. So we add them together.

packages/reactotron-app/App/Stores/DevTrackerStore.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @computed
+    get errorsPerSecond() {
+        const currentConnectionTime = this.connectionTime
+
+        if (currentConnectionTime === 0) return 0
```
rmevans9: To `Infinity` and beyond! (aka avoiding EPS being `Infinity`)

packages/reactotron-app/App/Stores/DevTrackerStore.js
```diff
@@ -0,0 +1,77 @@
...
+
+        return this.currentErrors / currentConnectionTime
+    }
+
+    @action
+    connectionStarted = () => {
+        if (!this.sessionStarted) {
```
rmevans9: Right now every single time connections change this method is called. We don't want to reset our timer just incase it is just a second device that connected

jamonholmgren:
Looks awesome! Are you sure you want "errors per second" and not "errors per minute" or "errors per hour" ?

rmevans9:
I was considering that - per a second is a bit much. Per a minute or hour could be good. It would actually be easy enough to setup a calculation for each and then give the user the ability to change over what time period that they want to see.

rmevans9:
A quick bit of refactoring to the DevTrackerStore and we now are tracking in real time better. That also solved if Reactotron shuts down with active connections since it is storing the current amount of time connected every second (TODO: Consider throttling this a little bit). Also added two more computed methods (Errors Per Minute and Errors Per Hour). Next up is going to be starting on the UI and then loop back on seeing if there is a better way to check if a command is an error (there might not be but I have to at least try!)

rmevans9:
screen shot 2018-05-06 at 10 48 26 pm

Tab has been started.

jamonholmgren:
This is such a great idea.

PR #658, opened 2018-05-08
[WIP] Collapse State into one tab

rmevans9:
This does basically what was done for the RN tab but in a slightly different way. @skellock I am interested on your thoughts. I went with a slightly more code reuse way for the header so that we could standardize on that more easily. If you like the direction I can implement the rest of the tabs with the same header.

screen shot 2018-05-07 at 8 47 40 pm

screen shot 2018-05-07 at 8 47 49 pm


packages/reactotron-app/App/State/State.js
```diff
@@ -0,0 +1,100 @@
...
+      {
+        icon: "import-export",
+        text: "Snapshots",
+        name: "backups",
+      },
+    ]
+
+    return (
+      <div style={Styles.container}>
+        <Header selectedTab={ui.stateSubNav} tabs={tabs} onSelectTab={this.handleSelectTab}>
```
rmevans9: We could take this a step further and have a `Tab` component that you wrap in a `TabContainer` or something then just build this all up for you.

rmevans9:
Closing in favor of the Tabs implementation

PR #659, opened 2018-05-08
Tabs - Another approach

rmevans9:
This is to show another approach I thought of while working on moving the tabs around.


rmevans9:
👍 I will get the RN tab up to using this method tonight. I am not sure if we are ready to add the new "main" tab with the stats yet so it might make sense to get the RN tab updated and then merge this and loop back on the new tab for stats and other information on another PR.

skellock:
Good call.  Make we could land #653 with #606 so they can live together on our "main"/"home" screen?

rmevans9:
Sounds good to me. I will get this one ready to land then update 653 and 606 to live on a new tab. I might merge them into the same branch so I can just knock both out quickly and be ready to land em.

rmevans9:
This switches all of the headers minus Timeline over. Still need to change up the collapse arrow and move that to the menu

rmevans9:
@skellock That should wrap this PR up.

skellock:
👍

PR #660, opened 2018-05-08
♻️ Upgrades app dependencies + state refactoring

skellock:
Most of this codebase is 2 years old, so I'd like to update it a bit with what we've learned.

Unfortunately, it's pretty destabilizing to do this.

* upgrades to React 16.x
* upgrades to mobx 5.x
* removes `shouldComponentUpdate` from all command views
* adds a buffer so we batch updates when receiving high-volume bursts (this was locking the ui a bit previously)
* removes the need to know all commands at the state level

Next steps for this PR:

I need to change how state backups work.  Currently we're editing `payload.name` of the command.  In hindsight, this is weird.  

I think we need to split the incoming commands into 2 event categories.  Commands that hit the timeline (historical) and commands that don't (signals).  

For the ones that don't, we need to manage/transform them somewhere else (like our SessionState).


packages/reactotron-app/App/Commands/StateActionCompleteCommand.js
```diff
@@ -1,29 +1,27 @@
...
   name: {
     color: Colors.bold,
     margin: 0,
-    paddingBottom: 10
-  }
+    paddingBottom: 10,
+  },
 }
 
+@observer
```
rmevans9: Do we need to make all the commands observable? Is this because of the MobX upgrade?

rmevans9:
I am 100% on board on changing how commands work and splitting out to different types. I don't know what the right way to do it would look like but I feel like there is a better way then we do it now.

packages/reactotron-app/App/Commands/StateActionCompleteCommand.js
skellock: I'm not sure we do.  I think the answer to "will it ever change?" is no.

skellock:
Ok, @rmevans9, with that [last commit](https://github.com/infinitered/reactotron/pull/660/commits/0bc67972cd3c580cab806639f4a86e0eb0aacd16), this PR is ready.

Backups are now separate from the timeline.

I also started organizing the app state a bit.  I broke off the piece responsible for backups & restores.

In future PRs, I'll organize as I go too.

Time to clear out 2 years of tech debt!

PR #662, opened 2018-05-09
Connection Filtering

rmevans9:
This is basically a redo of the Connection Filtering. It ended up with a bunch of merge conflicts because of recent changes and we are looking at switching out the UI of it entirely so I pulled the backend (and the dialog) bits into a new branch off master and here it is!

Left to do:
 - [x] Add New "Home" tab
 - [x] Add Connection Selection to a tab on the new "Home" tab
 - [x] Add shortcut to open connection selection dialog
 - [x] Test each tab/subtab for how it handles connection filtering and find a plan for each that is needed.


rmevans9:
@skellock 

https://stackoverflow.com/questions/40561073/websocket-not-closed-on-reload-appreact-native?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

"You can't reload in production so this isn't a problem". Yup.

rmevans9:
Maybe related?

https://github.com/facebook/react-native/issues/18775

rmevans9:
@skellock I would be inclined to say that this isn't our problem but a bug in RN Android right now. I tried a few tricks to try and fix this issue on the Reactotron side but was not successful. I even tried to basically poll the connections every X seconds and close off ones that stopped responding but the extra connections never stopped responding... sooo ¯\\\_(ツ)_/¯

rmevans9:
@skellock put in a start for the status bar. Still gotta wire up some of the text and such but its a start. Let me know if it is headed in the direction you were thinking.

rmevans9:
New goodness. Last known issue: Android is terrible at killing connections in dev mode. So far I can't find good identifying information to be able to handle this on the `reactotron` side

rmevans9:
Demo finally back to working shape but it seems like Subscriptions are not working right anymore. Going to sort that out now.

skellock:
a85abd19-840b-419b-9ab2-19f297ffc37a_text

PR #663, opened 2018-05-11
[WIP] ✨ Remote control of clipboard

skellock:
Not a huge feature, but only the simulator has clipboard sharing.  This gives us device as well.

### DIFF
--- a/packages/reactotron-app/App/Stores/SessionStore.js
+++ b/packages/reactotron-app/App/Stores/SessionStore.js
@@ -144,6 +144,8 @@ class Session {
   handleCommand = command => {
     if (command.type === "clear") {
       this.commandsManager.all.clear()
+    } else if (command.type === "clipboard.value") {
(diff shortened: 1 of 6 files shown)


packages/reactotron-app/App/Stores/SessionStore.js
```diff
@@ -144,6 +144,8 @@ class Session {
   handleCommand = command => {
     if (command.type === "clear") {
       this.commandsManager.all.clear()
+    } else if (command.type === "clipboard.value") {
+      console.log('clipboard', command.payload.value)
```
rmevans9: We just want to console log the value? Not put it on the users computer clipboard?

skellock: ugh... no that's a mistake.  thx

skellock:
I'm not sure how useful this is in hindsight.... nobody asked for it.  I'll close it a resurrect it if need be.

PR #666, opened 2018-05-11
🐛 prevents crashes when we don't have stack frames with filenames

skellock:
Not sure how it's happening, but sometimes we don't have a filename property passed along with the stack frame.

This would kill the app.


skellock:
I love how this is #666

PR #672, opened 2018-05-23
Fix for a networking issue: response body replaced by blob metadata since RN 0.54

psegalen:
RN supports Blobs since 0.54, it changed some things in the networking stack, since then Reactotron networking plugin shows blob metadata instead of responses body in the "Response" tab of a request.

This can easily be reproduced by using this repro: https://github.com/psegalen/rn-android-blob-failure

This PR aims to fix this issue by getting the actual blob body when a server response has a "blob" type.

### DIFF
--- a/packages/reactotron-react-native/src/plugins/networking.js
+++ b/packages/reactotron-react-native/src/plugins/networking.js
@@ -111,7 +111,11 @@ export default (pluginConfig = {}) => reactotron => {
         // Disable reason: FileReader should be in global scope since RN 0.54
         // eslint-disable-next-line no-undef
         const bReader = new FileReader()
-        bReader.addEventListener('loadend', () => sendResponse(bReader.result))
+        const brListener = () => {
+          sendResponse(bReader.result)
+          bReader.removeEventListener('loadend', brListener)
+        }
+        bReader.addEventListener('loadend', brListener)
         bReader.readAsText(response)
       } else {
         sendResponse(response)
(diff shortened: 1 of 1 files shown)


packages/reactotron-react-native/src/plugins/networking.js
```diff
@@ -74,40 +74,51 @@ export default (pluginConfig = {}) => reactotron => {
...
-      try {
-        // all i am saying, is give JSON a chance...
-        body = JSON.parse(response)
-      } catch (boom) {
-        body = response
+      if (type === 'blob' && typeof FileReader !== 'undefined') {
+        // Disable reason: FileReader should be in global scope since RN 0.54
+        // eslint-disable-next-line no-undef
+        const bReader = new FileReader()
+        bReader.addEventListener('loadend', () => sendResponse(bReader.result))
```
rmevans9: Do we have to cleanup after ourselves for this listener or does it auto cleanup listeners?

rmevans9:
LGTM assuming the added listener gets auto cleaned up somewhere along the way. I am glad it was a fairly minor change to get this working!

skellock:
wow!  looks good, i'll test it out today!

packages/reactotron-react-native/src/plugins/networking.js
rmevans9: It appears `FileReader` inherits from `EventTarget` which I believe is this package:
https://www.npmjs.com/package/event-target-shim

If that is the case it should also provide a `removeEventListener` method. If that is case this might be as easy calling that with the correct arguments whenever it makes the most sense to.

skellock:
Thanks again!

psegalen:
Sorry about deleting your first review @rmevans9, wrong click followed by a 10PM bad decision of confirming without reading... :(

PR #691, opened 2018-06-09
Custom commands to single client

### LINKED ISSUE #687: Custom commands should send to 1 device
Currently if I send a custom command, it gets sent to every device.

Let's change that to only send to the active device.

This is in 2.0.0-beta.5.

rmevans9:
fixes #687 

If there is a better way to get at the observable on SessionStore from UIStore I would love to hear it but this worked for now!

### DIFF
--- a/packages/demo-react-native/App/Config/ReactotronConfig.js
+++ b/packages/demo-react-native/App/Config/ReactotronConfig.js
@@ -28,4 +28,6 @@ if (__DEV__) {
   Reactotron.clear()
 }
 
+Reactotron.onCustomCommand('test', () => console.log('This is an example'))
+
 console.tron = Reactotron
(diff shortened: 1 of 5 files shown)


packages/demo-react-native/App/Config/ReactotronConfig.js
```diff
@@ -28,4 +28,6 @@ if (__DEV__) {
   Reactotron.clear()
 }
 
+Reactotron.onCustomCommand('test', () => console.log('This is an example'))
```
skellock: did you want to `console.tron.log` this instead?  I'm ok if you don't want to.

rmevans9: Yeah, I can do that. Not used to the whole `console.tron` thing because I just slap `Reactotron` on `console.log` (which is super simple now thanks to you!). Updating now.

rmevans9:
Updated console.tron. Merging based upon @skellock approval.

PR #694, opened 2018-06-10
Ceil the width/height

### LINKED ISSUE #686: Android screen size was way to many decimal points.
image

Let's round that up to an integer.  Haha.

rmevans9:
fixes #686


skellock:
Thanks!

PR #696, opened 2018-06-10
:sparkles: add check all/uncheck all buttons

JulienUsson:
Add "Check all" and "Uncheck all" buttons in FilterTimelineDialog for better debug experience.
image


skellock:
Thanks for your contribution!  I'll be cutting a build tomorrow that will include this.

PR #729, opened 2018-07-08
Make timeline the default

### LINKED ISSUE #722: Make the timeline front-row centre
I recently introduced a "home" tab, and it's not working out the way I want.

I wonder if we can just remove it entirely?  Somehow move that information into the status bar at the bottom instead?

I talked to @rmevans9 about it and here's his thoughts:

image

I think i wanted to put it in place since we're going to have more stuff in there soon?  Perhaps I got ahead of myself.

rmevans9:
Fixes #722 
Might be worth throwing this in the config once that is in place but this fixes it for short term.


skellock:
thanks Rich!

PR #749, opened 2018-07-16
Add typings for reactotron-redux

henrikra:
I added typings for `reactotron-redux`. I know they are incomplete but I use only one function without params so this works for me and many other people. If someone wants to expand this later it will be easy now :)

### DIFF
--- a/packages/reactotron-react-native/reactotron-react-native.d.ts
+++ b/packages/reactotron-react-native/reactotron-react-native.d.ts
@@ -199,3 +199,8 @@ declare module "reactotron-react-native" {
 
   export { instance as default }
 }
+
+declare module 'reactotron-redux' {
+  import { ReactotronPlugin, Reactotron } from 'reactotron-react-native';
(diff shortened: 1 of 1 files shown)


skellock:
Heyo!  Thanks for the PR! 🎉 

I'm a reluctant to put this in the `reactotron-react-native` typings since this is for a plugin.

I *think* the right place for it is `reactotron-redux`.

Since everyone has different plugins, and those plugins decorate the same `reactotron` instance, it gets a little more complicated than I would like to type Reactotron properly.

Here's what I've been doing.  Let's call this `setup-reactotron.ts` for example.

```ts
import reactotron from "reactotron-react-native"

// start with the bare reactotron type definition
type VanillaReactotron = typeof reactotron

// extend the type with features bestowed by plugins
export interface Reactotron extends VanillaReactotron {
  // here you would place anything additional features that come from plugins
}
```

I hope that makes sense.  

Thanks a tonne again, sorry for rejecting the PR.  That always makes me feel like a villain. 😈

PR #752, opened 2018-07-18
Small tweaks to contribution doc

robinheinze:
Clarify a few things in the contribution guide, and link from the main README.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -2,7 +2,7 @@
 
 ### Dev Tools
 
-Make sure you've got the latest `yarn` (1.7.0) and the latest `node` (8.6.0).  There are some 
+Make sure you've got the latest `yarn` (1.7.0) and the latest `node` (8.6.0).  There are some
 issues with older combos of these.  =)
 
 Also if you encounter a problem with `electron 1.8.6` passing checksums, there was a bug in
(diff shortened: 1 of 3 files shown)


docs/contributing.md
```diff
@@ -19,17 +19,28 @@ git clone git@github.com:infinitered/reactotron.git
 cd reactotron
 ```
 
-Install `lerna` and run the setup script to install the dependencies & run tests.
+If you don't already have `lerna`, install it:
+
+```
+npm install --global lerna
```
skellock: You won't need to install this globally.  This should come in as a local dependency because it's in the `package.json` at the root of the project.  I'm thinking we remove this part.

robinheinze:
Except, when I re-cloned and DIDN'T globally install lerna after getting that error, I also got a clean `yarn install`... So I'm doubtful that it has anything to do with lerna at all

skellock:
And as a surprise to absolutely nobody, the module requiring `fsevents` is indeed called `sane`.

skellock:
In the past, whenever i've gotten into a wierd spot with dependencies, nuking your `yarn.lock` so that very deep dependencies can get refreshed has helped.

I'm certain that's bad advice... but I'm that's all I got.

If a transitive dependency is locked to ^1.1.3 and that contains a bug that gets patched, the lock file will prevent us from getting it (since the whole point is to pin versions of every depth level).

robinheinze:
Blowing away the `yarn.lock` seemed to fix the problem 🤷‍♀️ 

If you see any red flags in the new one, let me know. It seemed to be mostly minor version differences, but `yarn.locks` make my head hurt so I can't be sure

Also removed the `lerna` installation line in the markdown!

PR #756, opened 2018-07-20
[WIP] Add copy as curl to API logs

robinheinze:
Adds Copy as Curl to API response options. I used the existing "content-copy" icon... let me know if you want to use something different.

screen shot 2018-07-20 at 12 09 59 pm


packages/reactotron-app/App/Shared/CommandToolbar.js
```diff
@@ -153,6 +158,22 @@ class CommandToolbar extends Component {
       const text = apiToMarkdown(this.props.command.payload)
       clipboard.writeText(text)
     } catch (e) {
+      clipboard.writeText(body)
```
skellock: `body` isn't defined in this function, so we'd end up throwing a new exception.

packages/reactotron-app/App/Shared/CommandToolbar.js
```diff
@@ -153,6 +158,22 @@ class CommandToolbar extends Component {
...
+    event.stopPropagation()
+    const { command } = this.props
+    const { payload } = command
+    const body = dotPath('request.data', payload)
+
+    try {
+      const text = apiRequestToCurl(this.props.command.payload)
+      clipboard.writeText(text)
+    } catch (e) {
+      clipboard.writeText(body)
```
skellock: This would put the payload of the `post` or `patch` on the clipboard which is not what we want for this function. 

Probably best to have a empty catch instead.

Every time I see this empty catches, I think... oh ya... i need to add a error message or something... then i never do.

packages/reactotron-app/App/Lib/api-to-curl.js
```diff
@@ -0,0 +1,30 @@
...
+    try {
+      if (typeof data === "string") {
+        parsedData = JSON.parse(parsedData)
+      }
+      parsedData = JSON.stringify(parsedData)
+    } catch (e) {}
+    output.push(` -d '${parsedData}'`)
+  }
+
+  return output.join("")
```
skellock: I wonder if we'll need to do shell escaping?  For example, payloads will be in `json` which have `"` characters out the wazoo.  Should we escape them to be safe for the shell?

packages/reactotron-app/App/Shared/CommandToolbar.js
skellock: probably best to leave this empty too...

robinheinze: Doh, this was a failed merge conflict resolution. Will fix. Thanks for catching.

packages/reactotron-app/App/Lib/api-to-curl.js
robinheinze: Are ” safe when they’re wrapped in ‘ ?

skellock: I don't think so... i think it'll probably end look like this:

```
curl -d "{\"lulz\": \"you won't escape\"}"
```

or 

```
curl -d '{"lulz": "you won\'t escape"}'
```

i think?  Not entirely positive on the shell escaping rules.

robinheinze:
@skellock Fixed the catch blocks, but failed miserably at the shell escaping. I tried to adapt the function from that lib (https://github.com/xxorax/node-shell-escape/blob/master/shell-escape.js), but I never seemed to out put what I expect. I couldn't really grok what the regexes were trying to accomplish.

So i'll try to get back to it this week if I have time, but feel free to take it over if you know you can make it work.

skellock:
No worries!  Thank you @robinheinze.  I'll mark this as `wip` and whoever gets to it first wins.

🙇

packages/reactotron-app/App/Lib/api-to-curl.js
Krisztiaan: Apparently one could also just escape everything that is not alphanumeric, or the boundary quotes.

rmevans9:
Updating this PR for all the changes in the repo. I am going to look to wrap this up and land it this coming week.

scottwio:
Any movement on this one would be really helpful

rmevans9:
I am going to merge this as is as it worked in my quick tests it works ok. We can iterate on it and fix any specific issues that popup. I am sure there will be some as there are a lot of different use cases for API requests.

Issue #591, opened 2018-01-12
Add missing types


yordis:
Non of the plugins from `reactotron-react-native` are exported in the typescript types

skellock:
I'm still unclear on how to extend typescript types with additional features from multiple plugins.

I think we're going to have to do a DIY union.  I just don't know of another way.

ajmotika:
Is there any update on this? We want to use reactotron-redux and reactotron-redux-saga with our React Native app written in typescript.

skellock:
You can click the subscribe button to the right to track if there's any updates to this.

ajmotika:
I created typings to get it to work with my project.  Here they are below.  

```javascript
import Reactotron, {ReactotronPlugin} from 'reactotron-react-native';
import {StoreCreator} from 'redux';

// Module Augmentation for plugin
declare module 'reactotron-react-native' {
  export interface Reactotron {
    createStore: StoreCreator;
  }
}

declare module 'reactotron-redux' {
  export function reactotronRedux(): (reactotron: Reactotron) => ReactotronPlugin;
}
```

rodrigoelp:
Hey guys, just fixed a small issue with the type definition in #748. I got no idea why the compiler didn't give you a warning before for `close()` but not a big issue.

Over the next few days will try to get some of the plugins to work

rodrigoelp:
Hi guys, to continue to extend the typescript definition. Here I included another bit of the definition: #757

rodrigoelp:
Hi @ajmotika, are you going to create a pull request with your type definitions? I am adding these little by little and got to the point you might want to contribute? otherwise, I can create the pull request and push it up for review.

ajmotika:
Thanks @rodrigoelp! if you have the PR ready to go then you can push it up for review. If not, I can relook at everything and get one up tomorrow.

rodrigoelp:
Hi @ajmotika, I did create a branch for it and just amended it to include a mention to your contribution (if you are happy with that). Ironically, somebody had created the typedef for redux but it was not linked to the `package.json` file so the type was never discoverable to typescript. I am just waiting for the approval of my PR to chuck this one in.

Alternatively, you can provide comments to the branch as I have pushed it up here: [redux branch](https://github.com/rodrigoelp/reactotron/tree/redux)

It includes the redux and redux-saga typedef.

rodrigoelp:
[...]

PR #757, opened 2018-07-21
Included extra plugin definitions of "reactotron-react-native" for typescript

rodrigoelp:
This commit enables the definition of the following scenario:

```typescript
import * as Reactotron from "reactotron-react-native";

Reactotron.default
    .configure({ name: config.name })
    .use(Reactotron.trackGlobalErrors()) // tracking errors
    .use(Reactotron.asyncStorage()) // tracking storage
    .use(Reactotron.networking()) // tracking networking
    // the lines above could be substituted by
    //.useReactNative() 
    .connect();
```

### DIFF
--- a/packages/reactotron-react-native/reactotron-react-native.d.ts
+++ b/packages/reactotron-react-native/reactotron-react-native.d.ts
@@ -195,6 +195,28 @@ declare module "reactotron-react-native" {
     benchmark(title?: string): ReactotronBenchmark
   }
 
+  /**
+   * Provides async storage information to Reactotron.
+   */
+  function asyncStorage(): (tron: Reactotron) => ReactotronPlugin;
+  /**
+   * Provides a global error handler to report errors.
+   */
+  function trackGlobalErrors(): (tron: Reactotron) => ReactotronPlugin;
+  function openInEditor(): (tron: Reactotron) => ReactotronPlugin;
+  /**
+   * Provides an image.
+   */
+  function overlay(): (tron: Reactotron) => ReactotronPlugin;
+  /**
+   * Provides network information to reactotron.
+   */
+  function networking(): (tron: Reactotron) => ReactotronPlugin;
+  /**
+   * A plugin which provides .storybookSwitcher() on Reactotron.
+   */
+  function storybook(): (tron: Reactotron) => ReactotronPlugin;
+
   var instance: Reactotron
 
   export { instance as default }


rmevans9:
I am going to let @skellock look at the code here as he does a lot more TypeScript then I do but just wanted to mention that when using `.useReactNative` it turns on all of those plugins you are showing there automatically. Not sure if you put it that way for an example or not so just wanted to make sure you knew.

rodrigoelp:
Hi @rmevans9, I added it by mistake. In my test, however, I had it off.

I will wait for @skellock to provide his comments. I might soon take in the type definitions the suggestion provided by @ajmotika in #591 if he/she does not create the pull request. On that pull request, I might just push as well the type definitions for redux-saga

skellock:
Ya, historically we always had the individual plugins as you've typed here, but @rmevans9 made a nice central one called `useReactNative`.  [This function](https://github.com/infinitered/reactotron/blob/master/packages/reactotron-react-native/reactotron-react-native.d.ts#L52) is fully typed and ready to go today.

For example: 

```js
Reactotron
  .configure()
  .useReactNative({
    asyncStorage: false,
    networking: {
      ignoreUrls: /symbolicate/
    }
  })
  .connect()
```

Thank you kindly for your PR ❤️ !  I always feel horrible when I'm not able to land new contributor's hard work... 😿 

Hopefully this makes sense.  Lemme know if I'm off my rocker.

skellock:
Also, @rmevans9 pointed out (offline) that whether or not we like the individual plugins, they are still part of the public interface.

So should we still type them?

I vote no... but i totally get that point.  I'd be ok with opening this back up and accepting it... then deprecating it for 3.0.

Whatever the concensus is... i'm ok.

rodrigoelp:
Hi @skellock, I am not particularly attached to this code, however, it is weird the typedef does not allow you to implement what the documentation states [here](https://github.com/infinitered/reactotron/blob/master/docs/quick-start-react-native.md#configure-reactotron-with-your-project).

In that case, fix the documentation to remove those references (if you are getting ready to deprecate these methods) as the documentation does not state (at least not in the getting ready) how to use the example you provided. What I am trying to say is, types are part of the documentation and in this particular case, the actual documentation and the types do not match, giving you the sense that the definition is broken.

Either way, I will be creating a new pull request to submit the redux and redux-saga components.

skellock:
Ya, you're totally right.  That's my fault.  I've been neglecting the docs.

#763 will right this injustice.

Sorry for the confusion.  :blush:

PR #759, opened 2018-07-22
Safely Import Reactotron

geirman:
Because Reactotron is a dev dependency, we want to be careful not to accidentally import the config in production. Therefore, wrapping the import in a conditional include makes sense as best practice.

### DIFF
--- a/docs/quick-start-react-native.md
+++ b/docs/quick-start-react-native.md
@@ -56,7 +56,9 @@ Finally, we import this on startup in
 on line 1:
 
 ```js
-import './ReactotronConfig'
+if(__DEV__) {
+  import('.ReactotronConfig').then(() => console.log('Reactotron Configured'))
+}
 ```
 
 At this point, Reactotron is hooked up.


geirman:
Apparently, according to [this SOF Answer](https://stackoverflow.com/a/46543835/1349269), it's part of the [babel-preset](https://babeljs.io/docs/en/babel-plugin-syntax-dynamic-import/) and is currently in Stage 3. And seems to be working well so far as I can tell.

I haven't been using the tron logs (yet). I can see how your suggestion makes sense if Reactotron weren't installed as a devDependency, but since it is, it seems like it makes sense to ensure that it only exists when in `__DEV__` mode by default.

skellock:
Thanks for the heads up dude.  I'll take a closer look at this.

How did you run into this issue btw?  Were you integrating this with CI and `devDependencies` weren't installed?  Or did you just see it and have your footgun sensors go off?

geirman:
Ha, I'm just trying to avoid the footgun thing. I realized I was importing a devDependency without any guards and figured that would probably blow up in my face if I ever forgot to comment out the import prior to deploying. I'm actually not sure what would happen in that scenario, but with the guard I'm pretty confident nothing bad will happen now.

PR #764, opened 2018-07-23
Fixed redux typedef issue and added redux-saga typedef

rodrigoelp:
Fixed small issue with redux's typedef as an interface should not define the accessor (everything on an interface is public via definition).

Redux-Saga type definition is included here with the augmentation to create the saga monitor.


rodrigoelp:
Sure, I can do that right now... no problem :)

PR #766, opened 2018-07-24
change subscribe to watch live state instead of persisted state unless…

kevinvangelder:
… a flag is used

We may also want to consider adding another config option that would allow globally hiding keys in subscription (such as password) that would only be displayed if the user clicks on a Show button.

### DIFF
--- a/packages/reactotron-mst/src/reactotron-mst.ts
+++ b/packages/reactotron-mst/src/reactotron-mst.ts
@@ -296,13 +296,14 @@ export function mst(opts: MstPluginOptions = {}) {
      * data within the state that will be sent to them every time it changes.
      *
      * @param command The command received from the reactotron app.
+     * @param subscribeToSnapshot Optionally subscribe to the persisted value instead of the live value
      */
-    function subscribe(command: any) {
+    function subscribe(command: any, subscribeToSnapshot?: boolean) {
       const trackedNode = trackedNodes[command.mstNodeName || "default"]
       const paths: string[] = (command && command.payload && command.payload.paths) || []
       if (trackedNode && trackedNode.node && paths) {
         subscriptions = uniq(flatten(paths))
-        const state = getSnapshot(trackedNode.node)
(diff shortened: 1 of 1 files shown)


skellock:
Thanks for following up on our conversation!  I'd love the feature and I'm ok with it being default 🎉 .

### Promote To Plugin Option

`subscribe` is internal. Placing the option `subscribeToSnapshot` is not accessible from the outside.

Instead, let's set the property on the [plugin options](https://github.com/infinitered/reactotron/blob/master/packages/reactotron-mst/src/reactotron-mst.ts#L89).

### Query Mode

This new mode can apply to [`requestKeys`](https://github.com/infinitered/reactotron/blob/master/packages/reactotron-mst/src/reactotron-mst.ts#L319) and [`requestValues`](https://github.com/infinitered/reactotron/blob/master/packages/reactotron-mst/src/reactotron-mst.ts#L338) too!

`subscribe`, `requestKeys`, and `requestValues` are all queries,  so let's call this thing `queryMode`.

```ts
export interface MstPluginOptions {
  /**
   * When requesting keys, values, or subscribing, configures whether
   * we talk to the live state object (great for `volatile` state) or the 
   * snapshot.  Defaults to `live`.
   */
  queryMode?: 'live' | 'snapshot'

  /**
   * Fine-grain control over what gets sent to the Reactotron app.
   */
  filter?: MstPluginFilter
}
```

### Tests

`reactotron-mst` is the newest Reactotron plugin, so it has the advantage of having unit tests (and written in ts).  Let's add some in for this feature.


### Docs

Let's add a small note about this to the [docs](https://github.com/infinitered/reactotron/blob/master/docs/plugin-mst.md#options) too. Also add a note that this doesn't apply to state backups.

Thanks!

kevinvangelder:
@skellock Updated as requested, but unfortunately I'm not able to run tests at the moment and I'm not really sure the right way to test this feature.

PR #772, opened 2018-08-03
Handling of gzipped/bytearray content in API Responses

Boris-c:
Hi Reactotron!

React Native on Android does not automatically uncompress gzipped API responses. This forces you to use `bytearray` as a `responseType`.

```
if (compress) {
    headers.push(['Accept-Encoding', 'gzip']);
    xhttp.responseType = 'arraybuffer';
} else {
    headers.push(['Accept-Encoding', 'identity']);
    xhttp.responseType = ''; // Defaults to text
}
```

image

This scenario is not currently supported by Reactotron (Response shows "skipped" instead).

This PR fixes that.

Before
image

After:
image

### DIFF
--- a/packages/reactotron-react-native/package.json
+++ b/packages/reactotron-react-native/package.json
@@ -69,6 +69,7 @@
   },
   "dependencies": {
     "mitt": "^1.1.2",
+    "pako": "1",
     "prop-types": "^15.5.10",
     "reactotron-core-client": "^2.0.0"
   }
--- a/packages/reactotron-react-native/src/helpers/byteArray-helpers.js
+++ b/packages/reactotron-react-native/src/helpers/byteArray-helpers.js
@@ -4,7 +4,7 @@ const pako = require("pako")
 
 export default {
   byteArrayToStr: (byteArray, method = "auto") => {
-    let actualMethod
+    let actualMethod = method
     if (method === "auto") {
       if (byteArray.length > 1) {
         // cf. http://www.zlib.org/rfc-gzip.html
--- a/packages/reactotron-react-native/src/helpers/byteArray-helpers.js
+++ b/packages/reactotron-react-native/src/helpers/byteArray-helpers.js
@@ -16,7 +16,7 @@ export default {
           actualMethod = "utf-8"
         }
       } else {
-        actualMethod = method
+        actualMethod = "utf-8"
       }
     }
     switch (actualMethod) {
--- a/packages/reactotron-react-native/src/helpers/byteArray-helpers.js
+++ b/packages/reactotron-react-native/src/helpers/byteArray-helpers.js
@@ -1,0 +1,30 @@
+import { decode as utf8Decode } from "./utf8-helpers"
+
+const pako = require("pako")
+
+export default {
+  byteArrayToStr: (byteArray, method = "auto") => {
+    let actualMethod
+    if (method === "auto") {
+      if (byteArray.length > 1) {
+        // cf. http://www.zlib.org/rfc-gzip.html
+        const [ID1, ID2] = byteArray
+        if (ID1 === 31 && ID2 === 139) {
+          actualMethod = "gzip"
+        } else {
+          actualMethod = "string"
+        }
+      } else {
+        actualMethod = method
+      }
+    }
+    switch (actualMethod) {
+      case "string":
+        return utf8Decode(byteArray)
+      case "gzip":
+        return pako.inflate(byteArray, { to: "string" })
+      default:
+        throw new Error(`Invalid byte array to string method: ${method}`)
+    }
+  },
+}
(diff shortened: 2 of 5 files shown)


packages/reactotron-react-native/package.json
```diff
@@ -69,6 +69,7 @@
   },
   "dependencies": {
     "mitt": "^1.1.2",
+    "pako": "1",
```
rmevans9: Are we sure RN doesn't ship with something like this already? I get the need for the lib but I just want to make sure we are not adding it just to add it.

Boris-c: As far as I know, RN doesn't ship with a zlib library, and pako seems to be the gold standard (3 million downloads last week).

Boris-c:
Interestingly, I had the exact same thought yesterday. So what I did is:
- find other JavaScript UTF-8 decoders (http://www.onicos.com/staff/iz/amuse/javascript/expert/utf.txt and `string_decoder` from https://www.npmjs.com/package/string_decoder)
- find test files (https://www.w3.org/2001/06/utf-8-test/UTF-8-demo.html, https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt and the longest Wikipedia article in Chinese, which funnily is about Lenin)

The goal was to test each of the current & other decoders for robustness and speed (on an iOS device).

My findings were that `string_decoder` is the fastest by a factor of 2-3, but it requires 2 additional dependencies (`buffer` and `safe-buffer`, i.e. that's 3 additional dependencies with `string_decoder` itself). `utf.txt` provides a bit faster decoder, but it fails with advanced/buggy UTF-8 even if it's fine with simpler strings. The code provided in this PR was as robust as `string_decoder`.

The decoding code with `string_decoder` is obviously much simpler:

```
static decodeWithDecoder(data) {
    const decoder = new StringDecoder('utf8');
    return decoder.write(Buffer.from(data));
}
```

I also tested the encoding codes, but that's not relevant to this PR.

By the way, in the decoding code I submitted here, you can gain a 10-20% performance increase by replacing all the `this.inRange` calls with the `min <= a && a <= max` equivalent.

Boris-c:
Hi, I'm sorry, I had to fix a few things in the helpers. I rushed maybe a bit before requesting the PR, it's good that you waited before merging ;)

I ran tests with the files I mentioned above, but both zipped and not, and things are fine now.

Boris-c:
I've also now modified a bit less `// prepare the right body to send` section and augmented the `sendResponse` function. As a result, gzipped text contents like HTML also show in Reactotron.

image

image

rmevans9:
Looks good me me @Boris-c. Any other changes you need to get in prior to merging?

@skellock any reservations with getting this in?

Boris-c:
@rmevans9 I've been using it as is for a few days, all API calls gzipped, both on iOS and Android, without any further changes to make on my side.

skellock:
Alrighty... lemme try this out then on my current project.  I don't believe the server is using gzip, so if this doesn't break it, I'd say, we're good.

skellock:
Are you passing `Accept-Encoding: gzip` to your API?  Apparently we shouldn't be in React Native.  React Native uses OkHttp under the covers in Android, and that handles the gzip decompression transparently.

Boris-c:
Ok, I investigated a bit and here are my observations and conclusions, after playing with the client code and watching the headers received by the server.

The frameworks add `Accept-Encoding: gzip` to the headers, when no `Accept-Encoding` is specified (i.e. on iOS: https://stackoverflow.com/questions/2682483/nsurlconnection-nsurlrequest-gzip-support). In that case, if the server returns a gzipped content, the latter is uncompressed behind the scenes and available for direct use (to the app and, obviously, to Reactotron).

If, on the other hand, you specify `Accept-Encoding: gzip`, then things get the way I mentioned in my first message: response is not uncompressed on Android, and you need to request `xhttp.responseType = 'arraybuffer'`, do the arraybuffer -> uncompressed -> UTF-8 yourself, which was the point of this PR.

My mistake was I did not know nor expect that `Accept-Encoding: gzip` could be added automatically. Second mistake, I couldn't imagine that the behavior would not be the same if I added it myself.

My conclusion is that you're right, things can work in the vast majority of cases without integrating this PR.

I leave it to you to decide whether Reactotron should handle the case when someone adds `Accept-Encoding: gzip` and requests an array buffer, either because he needs to or because he doesn't know gzip is automagically requested and managed by default.

For the sake of completeness, here are the request headers on iOS (similar on Android) when gzip is not requested :
image

Of course, Reactotron reports the headers added by me only:
image

In any case, thanks for the time you devoted to this PR, and big round of applauses for Reactotron ;)

skellock:
I agree with your assertion that this is not expected at all.  It seems like this is by design from the OkHttp team.  

I think the right approach here is to close the lid and pretend we saw nothing.

I really appreciate the detective work you've done!  I feel like I understand much better how to the compression stuff works now because of you.

Sorry we couldn't land this.

Boris-c:
I understand ;)

OkHttp calls that witchery `transparentGzip`...

image

https://github.com/square/okhttp/blob/33d31d073ee451a3848192a4737cc06ab83cbfd6/okhttp/src/main/java/okhttp3/internal/http/BridgeInterceptor.java

Issue #782, opened 2018-08-20
RFC: Drop Electron


skellock:
## Overview

I'd like to drop Electron and replace it with a local `Node.js` server.

_hold for applause_

## Why?

* Projects can include the reactotron server as a devDependency and type `yarn run reactotron`!
* The binary is a 154 MB dependency!
* It simplifies deployment from a maintainer point-of-view.
* Even though it's just a react-js app, I've heard from several people Electron is a barrier to contributing.
* There's nothing about Reactotron that uses any advanced Electron features.

## What do we lose?

Feature-wise?  Nothing.

A node-js server (like express, koa, etc) can still read/write files and clipboard.  Browsers can drag & drop file uploads for image overlays.

For people who like to run an app instead of a local webpage, they can still use https://github.com/jiahaog/nativefier to wrap it.

## How much work is this?

A new `reactotron-server` package would have to be created and configured to dish up a SPA. 

We'd also have to add endpoints for each write operation (sending state, commands, etc).  These would be merely pass thru to existing code.

The image overlay would be the only one I don't know how that works.  I've never done that before.  I suspect there's great React components that will do this just fine.

Keyboard shortcuts would have to be changed from using the `command`/`win` key.

## Current State of Electron App

The current Reactotron works great, but inside it's ugly.  It's about 2 years old and i literally learned react-js. I'd love to modernize this.  

We don't have to do this step now, but it might be worth considering since we're in the area anyway.

## Summary

This feels like the right step for Reactotron 3.

What do y'all think?  Please talk me out of this  🙏 .  Or push me 😈 .

ianoshorty:
### Context
I'm a multi-platform developer, full stack, working on everything from backend to frontend, desktop to mobile. I'm an active React Native dev, and use Reactotron most days. I'm not a contributor on the project.

### Opinion
I believe integrating reactotron into the RN development lifecycle, in a fashion similar to [Storybook](https://github.com/storybooks/storybook) would be incredibly useful.

- It guarantees all members of the team I'm working on my project with have the same debugging environment I do.
- I can version control that debugging with the project - ensuring maximum compatibility.
- I can run one command to boot and dev with my entire environment (e.g. `npm run reactotron`).
- I can limit all my application debugging to dev dependancies.
- I know there is no "pollution" of my debugging environment to other projects, where it may not be relevant.

I would be a big fan of this change. I'm also assuming this would reduce memory usage dramatically?

GantMan:
### Context
[...]

PR #787, opened 2018-08-22
[WIP] Working on dropping electron & picking up GraphQL

rmevans9:
This is no where near done. This is just the first set of changes that allows basic storage of connections and commands into GraphQL. There are many many many many many things to come. Putting this up here to start reviewing the direction.

Ref: https://github.com/infinitered/reactotron/issues/782


packages/reactotron-server/src/server/reactotron/commandHandler.ts
```diff
@@ -0,0 +1,11 @@
+import { commands } from "../datastore"
```
rmevans9: This and `connectionHandler` job should always be simple. Take a `reactotron` event and:

- Put it in the datastore
- Publish an event if needed

packages/reactotron-server/src/server/reactotron/connectionHandler.ts
```diff
@@ -0,0 +1,19 @@
+import { connections } from "../datastore"
+import { messaging, MessageTypes } from "../messaging"
+
+function onConnectionEstablished(connection) {
+  const newConnection = { clientId: connection.clientId, platform: connection.platform }
+  connections.addConnection(newConnection)
```
rmevans9: `addConnection` does some logic that decides if we should actually add the connection or not. Not sure if we should skip the event on the next line if it doesn't. Right now I don't believe it will cause any harm

packages/reactotron-server/src/server/reactotron/connectionHandler.ts
```diff
@@ -0,0 +1,19 @@
...
+  messaging.publish(MessageTypes.CONNECTION_ESTABLISHED, { connectionsUpdated: connections.all() })
+}
+
+function onConnectionDisconnected(connection) {
+    // TODO: Remove this junk
```
rmevans9: I remove the connection now. So it is time to remove this comment.

PR #791, opened 2018-08-23
GraphQL with 100% more type sharing. Also ability to execute commands via a mutation. NBD

rmevans9:
This introduces a nice little library that lets us type once and use twice!


rmevans9:
I have added the ability to execute commands on the reactotron client via GraphQL mutations. This is done in a way that the caller to GraphQL can pass pretty much everything it wants or needs and that is just shipped off to reactotron. We also track executed commands which could be used for a command history in the UI if we so please. Also added a subscription for it similar to the `commands` one which is for incoming commands from reactotron.

What does all that mean? This. NOTE: Does not depict the subscription

packages/reactotron-server/src/server/apollo/resolvers/commands.ts
```diff
@@ -0,0 +1,20 @@
+import { Resolver, Query, Subscription, Root } from "type-graphql"
+
+import { MessageTypes } from "../../messaging"
+import { Command } from "../../schema"
+import { commands } from "../../datastore"
+
+@Resolver()
+export class CommandsResolver {
```
rmevans9: I am enjoying the splitting of the resolvers. Makes everything much more management IMO

packages/reactotron-server/src/server/apollo/resolvers/executedCommands.ts
```diff
@@ -0,0 +1,46 @@
...
+      date: new Date(),
+      payload,
+    }
+
+    executedCommands.addCommand(newCommand)
+    messaging.publish(MessageTypes.EXECUTED_COMMAND_ADDED, newCommand)
+
+    // Welcome to the hack zone. Where everything is permitted as long as it works.
+    // TODO: Better name? Put this into a singleton that we can import and use instead of this function? Better everything? something? ANYTHING?
+    const reactotronServer = createReactotronServer()
```
rmevans9: I want to change how I handle the reactotron server but this serves the purpose for now. Note that this function returns a singleton so we are not actually re-creating the server every time this is called despite the name of the function.

packages/reactotron-server/src/server/datastore/connections.ts
```diff
@@ -1,33 +1,4 @@
-export type Connection = {
```
rmevans9: NO DUPLICATED TYPES. WOOOOOO

packages/reactotron-server/src/server/schema/command.ts
```diff
@@ -0,0 +1,26 @@
+import { ObjectType, Field, Int } from "type-graphql"
+import * as GraphQLJSON from "graphql-type-json"
+
+@ObjectType()
+export class Command {
```
rmevans9: So many decorators... but this makes me type once and use twice. Its glorious.

packages/reactotron-server/src/server/schema/connection.ts
```diff
@@ -0,0 +1,88 @@
+import { ObjectType, Field, Int } from "type-graphql"
+
+@ObjectType()
+export class Connection {
```
rmevans9: Making this a more complex object is on my hit list (In this case a more complex object means a simpler object for the consumer. I plan to combine some things into sub types so this isn't so flat)

packages/reactotron-server/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
rmevans9: I still need to work out a good solution to this. The build uses the other config but VSCode uses this one. I kinda don't like how that works but maybe its the way it needs to be? ¯\\\_(ツ)\_/¯

rmevans9:
Command filtering on both the query and subscription are now in. 

GraphQL is way too easy to reason with and build with. SOMEONE SHOULD PUT A STOP TO THIS MADNESS. IT IS MAKING ALL OTHER API BUILDING LOOK BAD.

Because I have so much extra time since building this is a cake walk have another gif.

packages/reactotron-server/src/index.html
```diff
@@ -0,0 +1,8 @@
+<html>
+    <head>
+        <title>This is gunna be AWESOME</title>
+    </head>
+    <body>
+        One day in the (hopefully) not far future this will be the new home of reactotron.
```
ianoshorty: 👍

packages/reactotron-server/src/server/apollo/index.ts
```diff
@@ -1,13 +1,21 @@
...
 
-import typeDefs from "./schema"
-import resolvers from "./resolvers"
+import { messaging } from "../messaging"
+import { resolvers } from "./resolvers"
 
 let apolloServer: ApolloServer
 
-export function createApolloServer() {
+export async function createApolloServer() {
```
ianoshorty: Perhaps it would be good to establish a convention for returning singletons? Something like `getApolloServer` or just `apolloServerInstance`?

ianoshorty: For context, I like the way Objective-C does this -> `SomeClass *instance = [SomeClass sharedInstance]`

packages/reactotron-server/src/server/apollo/resolvers/commands.ts
```diff
@@ -0,0 +1,26 @@
+import { Resolver, Query, Subscription, Root, Args, Arg } from "type-graphql"
+
+import { MessageTypes } from "../../messaging"
+import { Command } from "../../schema"
+import { commands } from "../../datastore"
```
ianoshorty: I can see a risk of an editor mistake being made here, having `Command` and `commands`. Perhaps this could be renamed to be a little more distinct?

packages/reactotron-server/src/server/apollo/resolvers/executedCommands.ts
ianoshorty: I've made a similar comment at the top about Apollo server. A standard for the project may help.

packages/reactotron-server/src/server/apollo/index.ts
rmevans9: I was thinking about that as I was wrapping up last night. I was also wondering if we need some sort of wrappers around ApolloServer or the Reactotron server to attach methods that are useful from the server side. I will be playing around with ideas here.

packages/reactotron-server/src/server/apollo/resolvers/executedCommands.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    reactotronServer.send(type, payload)
+    // Now leaving the hack zone.
+
+    return true
```
ianoshorty: I may be misunderstanding, so please forgive me, but what exactly does returning true imply here?

packages/reactotron-server/src/server/apollo/resolvers/commands.ts
rmevans9: Yeah, good call. I am thinking I will rename the `datastore` ones most likely

packages/reactotron-server/src/server/apollo/resolvers/executedCommands.ts
rmevans9: It implies that TypeScript/GraphQL (not sure which one right now...) wanted a return value. I need to look more into why that is.

packages/reactotron-server/src/server/apollo/resolvers/commands.ts
skellock: good call for sure!

packages/reactotron-server/src/server/apollo/resolvers/commands.ts
```diff
@@ -0,0 +1,26 @@
...
+  @Query(() => [Command])
+  commands(@Arg("clientId", { nullable: true }) clientId?: string) {
+    if (clientId) {
+      return commands.byClientId(clientId)
+    }
+
+    return commands.all()
+  }
+
+  @Subscription(() => Command, {
```
skellock: This is really starting to look like C#.  😆

packages/reactotron-server/src/server/datastore/connections.ts
skellock: saweeet!

packages/reactotron-server/src/server/schema/connection.ts
skellock: Ya, this is my fault.  I threw them all in there without any thought of organizing.

Can we leave them as is so we can try to ride out the current protocol for now?  One thing I love doing is using the existing reactotron in real projects, but swapping out the "app/backend" for the new version.

I understand we'll have to change the protocol, but I'd like to hold off as long as possible.

packages/reactotron-server/tsconfig.json
skellock: I wonder if there's a setting we can put inside `.vscode/settings.json` to clue VS Code in?

packages/reactotron-server/src/server/schema/connection.ts
rmevans9: My plan was to transform this object to a more complex one in the server here so `core-server` and `core-client` stay the same (for now). My hope is to not touch anything other then the new package so we are technically backward compatible for now.

packages/reactotron-server/src/server/apollo/resolvers/commands.ts
rmevans9: I am not complaining about that.

packages/reactotron-server/src/server/schema/connection.ts
skellock: great plan!  always 1 step ahead of me.

PR #795, opened 2018-08-25
[V3] Add ability to filter commands based upon command type (1 of 3)

rmevans9:
This lets you filter by command type in both query and subscription.

```
query {
  commands(filter: ["api.response"]) {
    type
  }
}
```

```
subscription {
  commandAdded(filter: ["api.response"]) {
    type
  }
}
```

We have not lost the ability to filter by clientId along with it.

```
query {
  commands(clientId: "My Client ID", filter: ["api.response"]) {
    type
  }
}
```


packages/reactotron-server/src/server/datastore/commands.ts
```diff
@@ -7,8 +7,22 @@ export class Commands {
...
-  byClientId(clientId) {
-    return this.commands.filter(cmd => cmd.clientId === clientId)
+  filterItem(item: Command, clientId?: string, filter?: [string]) {
+    return (
+      (!clientId || item.clientId === clientId) &&
+      (!filter || filter.some(fil => fil === item.type))
+    )
+  }
+
+  // Ideally this would be something we could figure out above but I couldn't get a boolean | Function return type to play well with the filter line below
```
rmevans9: @skellock maybe you have an idea here? Basically I wanted to make the command optional. If it was included I would evaluate to the boolean, if it was excluded I would return the function that you passed the command to which would then evaluate it. I got TypeScript to be happy about the function definition but the filter below on line 25 threw a fit because it didn't like the possibility of getting a boolean from the function call...

packages/reactotron-server/src/server/datastore/commands.ts
```diff
@@ -7,8 +7,22 @@ export class Commands {
     this.commands.push(command)
   }
 
-  byClientId(clientId) {
-    return this.commands.filter(cmd => cmd.clientId === clientId)
+  filterItem(item: Command, clientId?: string, filter?: [string]) {
```
rmevans9: With this function it becomes a little more then just a "store" because it offers the ability to run common filtering tasks against and arbitrary command. We should keep an eye on this and if this type of thing starts growing consider taking this (and the composable) function out of here and put it into a utility file or something.

rmevans9:
@skellock I think so a little bit. Here is what I wanted the signature to be:

```
filterItem(clientId?: string, filter?: [string], item?: Command) {
  if (!item) {
    return function(innerItem?: Command) {
      return this.filterItem(clientId, filter, innerItem)
    }

    return (
      (!clientId || item.clientId === clientId) &&
      (!filter || filter.some(fil => fil === item.type))
    )
  }
}
```

What this allows you to do is call this in two ways:

```
filterItem('Test', ['filter'], { type: 'test' }) // returns a boolean
```

and

```
const filterMe = filterItem('Test', ['filter']) // returns a function that is suitable for a .filter command
filterMe({ type: 'test' }) // returns a boolean 
```

rmevans9:
Merged in two other PRs because changes were getting all over the place. There are good comments in those merged PRs so worth a read through if you haven't already:

https://github.com/infinitered/reactotron/pull/797
https://github.com/infinitered/reactotron/pull/796

PR #796, opened 2018-08-26
[V3] plugin system (2 of 3)

rmevans9:
This introduces what could be the start of a plugin system to be used with V3.


packages/reactotron-server-plugin-redux/src/apollo/resolvers.ts
```diff
@@ -0,0 +1,13 @@
+import { Resolver, Query } from "type-graphql"
+
+import { TestSchema } from "./schema"
+
+@Resolver()
+export class TestSchemaResolver {
+  @Query(() => TestSchema)
+  testScehma() {
```
rmevans9: Yup... can't spell. Thats ok though. This is throw away code

packages/reactotron-server/src/server/pluginManager/index.ts
```diff
@@ -0,0 +1,65 @@
+import * as path from "path"
```
rmevans9: Let me start by saying that I am not sure if I like how this is laid out yet. I am throwing this out there as it is the first solution I landed on that works. Right now plugins can add schema, queries, subscriptions, etc to GraphQL. They can also provide event handlers for commands. These two things together can open up the ability to offload some functions from core like state snapshots. Why not put them in `reactotron-server-plugin-redux` or `reactotron-server-plugin-mst` since in core we don't really care. Not suggesting we dump subscription backups from core _yet_ but I want to open that door so we can slim core down to basically nothing _eventually_.

rmevans9:
Plugin development at this stage is pretty easy...

1. Open your plugin in a command line.
2. `npm link`
3. Open a command line over on `reactotron-server`
4. `npm link <PLUGIN NAME>`

I am hoping it stays that easy, if not easier.

rmevans9:
Things to consider:

- Dependencies in plugins. Right now the plugin will have (well, *have to* is kinda strong... but at this point I don't know how to do it without this) to depend on `type-graphql` which the server does too. The plugin should avoid bundling that in itself since the module should be present in the `node_modules` where it will reside
- This doesn't cover opening up the ability for plugins to participate in the UI _yet_. I have a few ideas of how that might work but will have to work with @skellock once he starts figuring out the frontend bit.
- We would probably want to work out how to build a plugin in plain ol JS as well since not everyone will want to use TS. (side note: do we actually care? We could put out a mandate that we use TS so we only help support plugins in TS? Seems like that might be heavy handed.)
- This does not currently provide access to the core datastores to plugins (commands, connections, executeCommands). Not sure if it ever will. I can't think of a reason to need that currently but I am sure there is one out there somewhere.
- As noted in the review comments - this forces plugins to take a specific name structure (`reactotron-server-plugin-<NAME>`). I think we will want to loosen that grip a little bit eventually but still try and keep it so we don't have to traverse _every single module_ in a project.
- My thoughts behind the plugin system is to try and offload absolutely everything we can off core. In the perfect world core would provide only a few things including the datastore, graphql server (that plugins can hook into in order to provide more information for the UI) and a really basic UI that gives a few main things like the timeline and such. Ideally it would have nothing specific to state management, api requests/responses, react native... or anything else that is not absolutely needed to work. Yes, I know this want basically means all the different bits we currently provide like `reactotron-redux`, `reactotron-redux-saga`, `reactotron-mst`, etc will need to be re-written. It might be that we keep this stuff in core but provide the ability to have them outside of core and work through each of the existing packages one by one in time. Once all of them are ready then pull stuff out of core that we can.

rmevans9:
Changed up a little bit and expanded the example to include a little bit to show of a client side (in the RN app) plugin to send information and have the server side plugin grab it and use it to populate a store which is available via GraphQL.

I don't particularly care for how the datastore feels in the plugin but I feel like not many plugins will actually need that and will instead simply provide UI to pull from the `commands` store and show data that way.

PR #797, opened 2018-08-28
[V3] backups on the server side (3 of 3)

rmevans9:
FWIW I dislike that I added this into core and am counting down the days (maybe weeks? eh... months probably) until we can yank this out and put it in a plugin. I put it in core for now so that we can keep parity a little easier with V2 until we can rework `reactotron-redux` and `reactotron-mst` to use the new plugin system.


rmevans9:
As I was typing the above comment it occurred to me that I could simply make a `reactotron-plugin-parity` or `reactotron-plugin-legacy` plugin that added these features until we get to adding them to where they actually belong. Current me is a genius and should teach previous me how to do things.

rmevans9:
I have moved the code to a plugin (woo!) and also to help solve dependency hell I introduced yarn workspaces. I don't know if this will work in the long run but it helped out with getting dependencies sorted for the plugins for now.

PR #798, opened 2018-08-28
Contributing documentation update.

ianoshorty:
Updates the documentation to references TS and TSLint as deps for contributing.


ianoshorty:
It occurs to me that this isn't technically a "clean" PR, due to the changes to `yarn.lock` files, however as they were automatically made during yarn install, I have included them. If they are not required, im happy to revert them and and update the PR.

rmevans9:
We will want to drop the yarn.lock changes here, at least the reactotorn-server one because it is "dropping" a dep that is required. I think its because I need to fix the package.json

ianoshorty:
@rmevans9 I've reverted that change, leaving just the documentation change.

PR #799, opened 2018-08-30
Refactor bootstrap and add tests

ianoshorty:
This PR ensures that the `reactotron-server` comes online and serves our index.html file.

I have done this mostly so I can get up to speed with whats going on in the project, but basically this PR:

- Refactors the index.js bootstrap process so that is unit-testable - separating out logic and only calling the "bootUp" method if the app isn't being executed in a test environment.
- Adds tests for the index.js file, to verify that an express server is booted correctly, and when accessed it serves out our dummy index.html file. This test checks to make sure some content is as we expect it.
- Creates a temporary script to copy index.html from the `src` directory to build / dist (as defined in the script). This is because TypeScript only compiles JS stuff, and we dont have webpack (or whichever bundle tool we intend to use) and I wanted to get the system working with tests passing.
- Adds a flag to the test config so that we dont get errors on unusedlocals. I agree that for actual production code this is useful, but for testing its annoying, especially when you're doing TDD. 

Hope this is useful / relevant.

### DIFF
--- a/packages/reactotron-server/src/server/index.test.ts
+++ b/packages/reactotron-server/src/server/index.test.ts
@@ -2,7 +2,7 @@ import test from "ava"
 import * as request from "request-promise-native"
 import { httpServerInstance, apolloServerInstance } from './'
 
-// Tests in here are ran in serial due to the webserver binding to port 4000 (concurrency results in EADDRINUSE)
+// Tests in here are ran in serial due to the webserver binding to port 4000 (concurrency results in EADDRINUSE error)
 
 let app, httpServer;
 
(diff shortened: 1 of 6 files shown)


packages/reactotron-server/src/server/index.test.ts
```diff
@@ -0,0 +1,36 @@
...
+test.serial("Can create a http server to serve react app", async t => {
+
+  let appServer;
+
+  t.notThrows(() => {
+    appServer = httpServerInstance()
+  })
+
+  // Not sure if there is a way to nicely destructure here, this does for now
+  app = appServer.app;
```
ianoshorty: If you guys have a good refactor for this, let me know! Scope wise I was running into an issue trying to get the app and httpServer properties within notThrows, but passing them out the function.

packages/reactotron-server/src/server/index.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  httpServer = appServer.httpServer;
+
+  t.true(httpServer !== null && typeof httpServer === "object")
+
+  const responseString = await request("http://localhost:4000").catch(err => {
```
ianoshorty: Ideally we would make the port option come from config here, but this is a fine first pass and we can update as we update the src.

packages/reactotron-server/tsconfig.test.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "sourceMap": false,
     "inlineSourceMap": true,
-    "outDir": "build",
+    "outDir": "build/server",
+    "noUnusedLocals": false,
```
ianoshorty: Calling this out just to make sure you guys are happy. Note this only applies in tests, not in actual source.

PR #808, opened 2018-09-06
Dynamic HTML for the website.

rmevans9:
The idea here is that instead of having a static file we dynamically generate it. This opens up the ability for the server to help out the UI with plugin resolution by figuring out where the plugin bundles are and adding them as scripts to the page. This would allow the plugin bundles to add things to the window for reactotron to use OR we could make it go the other way around where reactotron gets put on first and creates entrypoints that plugins can call into to register different things (like a tab they want to contribute to the app or a list of commands to add to the filter on timeline)


rmevans9:
@ianoshorty I agree with not pushing the 3rd party ui support any further yet. This was just a direction a tested for dynamically loading code into the webpage for use by the main application that I liked. As far as I am concerned we can leave the `getUiScripts` method as is (stubbed out returning nothing) since there is still a lot of open questions around 3rd party ui support that we need to answer before we can actually bring that feature to the finish line. 

If nothing else this PR removes the need to copy around an HTML file which I think works out for the better.

PR #811, opened 2018-09-07
Begins the web app

skellock:
Just adding dependencies at this point.

It isn't hooked up just yet to express.

To test:

```sh
yarn start:app
```

I added just enough to get started with this new fangled graphql thingy.

image


packages/reactotron-server/public/bundle.js
```diff
@@ -0,0 +1 @@
+console.log('yo')
```
rmevans9: yooooooo

skellock: yoooooooooo....  new console, who dis?

PR #814, opened 2018-09-07
✨ adds parcel bundler support

skellock:
This PR hooks us up with [parcel](https://github.com/parcel-bundler/parcel).

For dev, fire up two processes:

```sh
yarn start:server
yarn start:app
```

Our web app will now recompile & HMR when there are changes.

We're now hosted under `http://localhost:4000`.

This doesn't take into account how prod will get packaged, but we're not quite there yet, I reckon.

@rmevans9 apologies, I refactored the `bootUp()` code and the tests for it didn't survive.  I'll chop it up a bit more and add the tests back though.  Sorry dude. Hehe.  I'm the worst.


packages/reactotron-server/package.json
```diff
@@ -4,12 +4,12 @@
   "description": "A reactotron server",
   "main": "index.js",
   "scripts": {
-    "compile": "tsc",
+    "compile": "tsc -p tsconfig.server.json",
```
rmevans9: If we are building everything with the same config we can possibly just rename the config to be the default one

packages/reactotron-server/src/server/index.ts
```diff
@@ -16,52 +17,43 @@ const argv = argsParser(process.argv.slice(2), {
...
-  reactotronServer = null,
-) => {
-  // If we have been given a reactotron server instance, start listening
-  if (reactotronServer) {
-    reactotronServer.start(config.reactotronPort)
-  }
+  // attach some endpoints to power our web app
+  const URL_PATH_TO_PLUGINS = "/reactotron-plugins.js"
+  app.get("/", Endpoints.createReactotronAppEndpoint(URL_PATH_TO_PLUGINS))
+  app.get(URL_PATH_TO_PLUGINS, Endpoints.createInjectPluginsEndpoint(pluginManager))
```
rmevans9: I am not sure I follow how we are going to funnel multiple plugins scripts into a single endpoint like this? It might just be because I haven't tried it yet.

packages/reactotron-server/package.json
skellock: Good call.

packages/reactotron-server/src/server/index.ts
skellock: This will be a 2nd endpoint that the web app will hit upon loading.  The `/reactotron-plugins.js` endpoint will then visit all the plugins available and collect up the javascript strings needed to power the plugins.

Now, what happens next, I haven't quite thought through.

I suspect that's what you're asking about.  So incase you're wondering, my answer is:  dunno yet.  :laughing:

rmevans9: I think I came up with a few ideas over the weekend.

packages/reactotron-server/package.json
skellock: #817 fixes this.  Thx.

PR #818, opened 2018-09-10
✨ brings in some ui

skellock:
This brings in [tailwind](https://github.com/tailwindcss/tailwindcss) for some styling as well as a `<Text />` with `i18n` support.

For the future:

* I'd like for that `<Text />` component to be available to plugins... not sure how just yet.
* We're going to data-drive those tabs, so don't sweat that too much.  I just ran out of time for today.


rmevans9:
I am wondering if the components we want to share with the plugin should love in `reactotron-core-plugin` and then in the main app we could import them from there and use them. That way everyone gets to use the components.

rmevans9:
I guess the only thing the above suggestion doesn't handle well is the translations. That one is an interesting one but I think it is something we can handle by way of letting plugins provide their translations and we merge all the translations together before setting up i18n.

PR #823, opened 2018-09-18
[docs] Updates installing docs to explain how to open

m2mathew:
As an infrequent `brew ` user, I did not catch that adding the `cask` argument makes a package into an application. Since I spent a while looking around my computer for `reactotron`, I thought this small update might help the next person who was a little behind on the `brew` command landscape.

Thanks for making some fun tools to use, everyone!


rmevans9:
LGTM

PR #824, opened 2018-09-19
Remove i18n

rmevans9:


rmevans9:
Not sure if you were looking to remove it completely based upon the comment in the last PR but here we are removing it completely!

PR #828, opened 2018-09-20
Command clearing

rmevans9:
I have actually had this laying around for a while and just now realized it was not ever PR'd. So here it is.

### DIFF
--- a/packages/reactotron-server/src/server/datastore/commands.ts
+++ b/packages/reactotron-server/src/server/datastore/commands.ts
@@ -52,12 +52,17 @@ export function createCommandMatcher(clientId?: string, commandTypes?: [string])
 
 export class Commands {
   commands: Command[] = []
+  maxCommands: number
+
+  constructor(maxCommands: number = 300) {
+    this.maxCommands = maxCommands
+  }
 
   addCommand(command: Command) {
     this.commands.unshift(command)
 
-    if (this.commands.length > 300) { // Arbitrary, I know... but its a start somewhere
-      this.commands = this.commands.slice(0, 300)
+    if (this.commands.length > this.maxCommands) {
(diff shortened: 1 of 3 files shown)


packages/reactotron-server/src/server/datastore/commands.ts
```diff
@@ -54,7 +54,20 @@ export class Commands {
   commands: Command[] = []
 
   addCommand(command: Command) {
-    this.commands.push(command)
+    this.commands.unshift(command)
+
+    if (this.commands.length > 300) { // Arbitrary, I know... but its a start somewhere
+      this.commands = this.commands.slice(0, 300)
```
ianoshorty: Sometimes magic numbers are required, but if they are, I usually like to define them in a named constant so their purpose is clear. Its not obvious from this line why or what 300 is / does

skellock: I agree. I’d like this number to be configurable too. So maybe a constructor to pass it in? doesnt have to happen now though.

packages/reactotron-server/src/server/reactotron/commandHandler.ts
```diff
@@ -12,6 +12,12 @@ import { pluginManager } from "../pluginManager"
  * @param command A command sent from the client.
  */
 export function onCommand(command: Command) {
+  // TODO: Consider a special place for these "system" level commands. Do we want them to propagate like a normal command?
```
skellock: excellent point

packages/reactotron-server/src/server/datastore/commands.ts
rmevans9: Constructor pass in setup but didn't set up anything to actually pass it in. That part can come later.

PR #829, opened 2018-09-20
Allow for screens to have their own components for rendering

rmevans9:
This is the first step in allowing each of the screens to be built. From here we can build each screen and figure out how we want the screens object to be built up eventually. For now it is just hard coded which screens show up.

### DIFF
--- a/packages/reactotron-server/src/app/screens/Timeline/timeline-commands-list.tsx
+++ b/packages/reactotron-server/src/app/screens/Timeline/timeline-commands-list.tsx
@@ -1,0 +1,26 @@
+import React from "react"
+
+import { getCommand } from "./commands"
+
+interface Props {
+  subscribeToCommands: Function
+  commands: any[]
+}
+
+export class TimelineCommandsList extends React.Component<Props> {
+  componentDidMount() {
+    this.props.subscribeToCommands() // TODO: Do we gotta unsub?
+  }
+
+  renderCommand(command) {
+    const CommandComponent = getCommand(command.type)
+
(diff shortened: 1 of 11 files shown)


packages/reactotron-server/src/app/screens/Timeline/timeline-screen.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+`
+
+export class TimelineScreen extends React.Component {
+  render() {
+    return (
+      <div>
+        If everything works right you should see a lot of commands here:{" "}
+        <div>
+          <Query query={COMMAND_QUERY}>
```
rmevans9: This query takes a little bit when switching to the timeline page if there are a lot of events. Might need to look at only loading the last 10-20 or so and then allow for infinite scrolling or something of that nature.

packages/reactotron-server/src/app/screens/Timeline/timeline-commands-list.tsx
```diff
@@ -0,0 +1,26 @@
...
+  renderCommand(command) {
+    const CommandComponent = getCommand(command.type)
+
+    if (!CommandComponent) return null
+
+    return <CommandComponent command={command} key={command.messageId} />
+  }
+
+  render() {
+    return this.props.commands.map(this.renderCommand)
```
rmevans9: We might need to consider some sort of virtualized list.

rmevans9:
Merging in for now. Will address TODOs in time.

PR #831, opened 2018-09-23
Various fixes for getting started

ianoshorty:
Upon re-setting up the project for v3, I discovered some issues. These changes got me up and running. Includes:

- Ensuring reactotron-core-ui was included as a yarn workspace (and therefore `yarn install` runs automatically)
- Add missing proptype dec for text component
- Update contributing docs to help get going.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
   "workspaces": [
     "packages/reactotron-server",
     "packages/reactotron-core-plugin",
-    "packages/reactotron-plugin-*"
+    "packages/reactotron-plugin-*",
+    "packages/reactotron-core-ui"
   ],
   "lint-staged": {
     "subTaskConcurrency": 1,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -27,7 +27,8 @@
   "workspaces": [
     "packages/reactotron-server",
     "packages/reactotron-core-plugin",
-    "packages/reactotron-plugin-*"
+    "packages/reactotron-plugin-*",
+    "packages/reactotron-core-ui"
```
rmevans9: Crap, must of forgotten this. Nice catch!

packages/reactotron-core-ui/src/text/text.tsx
```diff
@@ -29,6 +29,7 @@ export interface TextProps {
   text?: string
   variant?: TEXT_VARIANTS
   className?: string
+  tx?: string
```
rmevans9: `tx` was actually removed so if we have packages using it we should dump those.

rmevans9:
Approving - I will cleanup the `tx` so we can get this merged in.

PR #832, opened 2018-09-23
Add Saga Timeline Componenet

ianoshorty:
Adds a saga timeline component in the same style as API response. Adds some basic styling to match existing app.

NOTE: This includes changes made in https://github.com/infinitered/reactotron/pull/831 so suggest reviewing that PR first.

### DIFF
--- a/packages/reactotron-server/src/app/screens/Timeline/commands/ApiResponse.tsx
+++ b/packages/reactotron-server/src/app/screens/Timeline/commands/ApiResponse.tsx
@@ -1,0 +1,13 @@
+import React from "react"
+
+interface Props {
+  command: any // TODO: Better typing yo
+}
+
+export class ApiResponse extends React.Component<Props> {
+  render() {
+    const { command } = this.props
+
+    return <div>{command.payload.request.url}</div>
+  }
+}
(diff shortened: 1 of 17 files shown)


packages/reactotron-server/src/app/screens/Timeline/commands/ApiResponse.tsx
```diff
@@ -8,6 +8,11 @@ export class ApiResponse extends React.Component<Props> {
   render() {
     const { command } = this.props
 
-    return <div>{command.payload.request.url}</div>
+    return <div className="bg-content border flex flex-row justify-between p-6">
```
rmevans9: Awesome! We need to work out a nice way to have the commands expandable similar to how they work in the app currently. I will see if I can make some way to make that happen in a shareable way.

packages/reactotron-server/src/app/sections/section-title.tsx
```diff
@@ -0,0 +1,29 @@
+// --- imports ---
```
rmevans9: Is this file used somewhere? I didn't see it anywhere in this PR.

PR #842, opened 2018-10-01
More typing and make scrolling work (maybe?) on timeline

rmevans9:


packages/reactotron-server/src/app/screens/Timeline/timeline-screen.tsx
```diff
@@ -34,13 +34,13 @@ export class TimelineScreen extends React.Component {
     return (
       <div className="flex flex-1 flex-col">
         If everything works right you should see a lot of commands here:{" "}
-        <div className="flex flex-1">
```
ianoshorty: Haha. I spent an hour trying to figure this out, sat befuddled. Well done dude, glad you got it sorted.

PR #844, opened 2018-10-01
cleans up the timeline

skellock:
I made the colours a bit more generic and adjusted some of the styling related to the timeline.

image

### DIFF
--- a/packages/reactotron-server/src/app/main/main.tsx
+++ b/packages/reactotron-server/src/app/main/main.tsx
@@ -1,6 +1,5 @@
 import React from "react"
 import ReactDOM from "react-dom"
-import { Text } from "reactotron-core-ui"
 import { Apollo } from "./apollo"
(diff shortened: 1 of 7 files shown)


packages/reactotron-server/src/app/main/main.tsx
```diff
@@ -1,6 +1,5 @@
 import React from "react"
 import ReactDOM from "react-dom"
-import { Text } from "reactotron-core-ui"
```
ianoshorty: I took this out, as I was getting compilation errors from Typescript with it in.

ianoshorty: To be clear, the error was simply that it was unused, but I couldnt build with it in.

skellock:
Ya, that makes sense.  Maybe build it on startup just to be safe.

PR #845, opened 2018-10-02
API Timeline Element Expansion

ianoshorty:
This PR modifies the APIResponse timeline element so that some of the detail is displayed and laid out. Im expecting some of the content here will be refactored into the shared ui component library, but for now it can act as a basis / demo. See below.

screenshot 2018-10-02 at 20 48 50

Next will need to be some layout control code for the JSON, but for now this is a start.

### DIFF
--- a/packages/reactotron-server/src/app/commands/ApiResponse.tsx
+++ b/packages/reactotron-server/src/app/commands/ApiResponse.tsx
@@ -1,32 +1,90 @@
 import React from "react"
 import { Command } from "reactotron-core-ui"
 
-export function ApiResponseTimeline({ command }: { command: Command }) {
-  console.log(command)
-  let contents = JSON.stringify(command.payload.request);
-  return <div>
-      <span className="text-highlight">{command.payload.request.url}</span>
+interface Props {
+  command: Command
+}
+
+interface State {
+  expanded: null | string
+}
+
+export class ApiResponseTimeline extends React.Component <Props, State> {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      expanded: null,
+    }
+  }
+
+  _renderContentsBlock() {
+    let contents
+
+    switch (this.state.expanded) {
+      case "response":
+        contents = JSON.stringify(this.props.command.payload.response.body)
+        break
+      case "responseHeaders":
+        contents = JSON.stringify(this.props.command.payload.response.headers)
+        break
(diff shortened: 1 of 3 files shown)


packages/reactotron-server/src/app/commands/ApiResponse.tsx
```diff
@@ -1,8 +1,88 @@
...
 import { Command } from "reactotron-core-ui"
 
-export function ApiResponseTimeline({ command }: { command: Command }) {
-  return <span>{JSON.stringify(command.payload.request)}</span>
+interface Props {
+  command: Command
+}
+
+interface State {
+  expanded: null | string
```
skellock: how about typing this instead?

```
expanded: 'none' | 'response' | 'responseHeaders' | 'requestHeaders'
```

Death to `null` and `undefined`!

packages/reactotron-server/src/app/commands/ApiResponse.tsx
```diff
@@ -1,8 +1,88 @@
...
+}
+
+export class ApiResponseTimeline extends React.Component <Props, State> {
+  constructor(props) {
```
skellock: This constructor could be replaced with a one-liner if you're interested.

```
state: State = { expanded: 'none' }
```

packages/reactotron-server/src/app/commands/ApiResponse.tsx
```diff
@@ -1,8 +1,88 @@
...
+    super(props)
+
+    this.state = {
+      expanded: null,
+    }
+  }
+
+  _renderContentsBlock() {
```
skellock: I prefer `private renderContentsBlock()` instead of the `_` prefix now that we have a way to represent visibility modifiers in this language.

packages/reactotron-server/tailwind.js
```diff
@@ -41,6 +41,9 @@ module.exports = {
   backgroundColors: {
     sectionPicker: colors.black,
     content: colors.greyDarkest,
+    expandButton: colors.greyDarker,
```
skellock: I like the extra roles... i wonder if we need the `expand` prefix at all though.  Will all buttons be like this?  Not sure. Now I'm thinking outloud.

ianoshorty:
Changes made. Love that 1 liner for the constructor. :-)

Issue #755, opened 2018-07-20
Reverse the log order


tjbenton:
Currently the log order is in reversed order from the way it it's written and it's very confusing to see it in reverse order especially when debugging complicated things.

It would be nice for this to be changed to match other consoles, or at least have a option for this.

![screen shot 2018-07-20 at 11 02 28 am](https://user-images.githubusercontent.com/4776422/43009730-6e8b1ea6-8c0c-11e8-9480-9cbd052ee29c.jpg)

in the chrome console this is what you get, and it's what I expected with Reactotron 

![screen shot 2018-07-20 at 1 11 58 pm](https://user-images.githubusercontent.com/4776422/43015821-a67a8434-8c1e-11e8-8219-3732a7834862.jpg)

anmead:
This would be great 👍

rmevans9:
https://github.com/infinitered/reactotron/blob/master/packages/reactotron-app/App/Lib/commands.js#L32

This is where the reversing happens. I would suggest that we would want some sort of toggle for this (maybe in the config file @skellock has been playing around with). Once the config is in place I am sure a PR would be accepted!

skellock:
There's a bunch of things hinging on that config file.  I just need to sit down and create it.  It'll take an hour.  (lol, still haven't learned my lesson about saying 💩 like that)

JulienMalige:
Can be closed :) @rmevans9 @skellock

rmevans9:
This has been released. Thanks!

PR #849, opened 2018-10-05
#755 Feature/reverse timeline

JulienUsson:
Hello,

This PR resolves issue #755 

Thanks,
Julien Usson

### DIFF
--- a/packages/reactotron-app/App/Stores/UiStore.js
+++ b/packages/reactotron-app/App/Stores/UiStore.js
@@ -26,7 +26,7 @@ class UI {
   @observable customMessage = ""
   @observable showSendCustomDialog = false
   @observable isTimelineSearchVisible = false
-  @observable isTimelineOrderReversed = false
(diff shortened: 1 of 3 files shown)


rmevans9:
This PR looks good to me. A note though: We are currently working on a new version of Reactotron (V3) in the V3 branch on here. It is a complete rewrite of the reactotron app (switches it to a web app). If you would like to help contribute to that we would love to have the help!

packages/reactotron-app/App/Stores/UiStore.js
```diff
@@ -135,6 +136,11 @@ class UI {
     }
   }
 
+  @action
+  toggleTimelineOrder = () => {
+    this.isTimelineOrderReversed = !this.isTimelineOrderReversed
```
rmevans9: It might be worth saving this so that you don't have to set it every time you open the app. Filters do something similar. I believe they store it into local storage.

JulienUsson: Done :)

JulienUsson:
Doesn't know about the V3, i'll give it a try :)

PR #851, opened 2018-10-07
Subscriptions server side

rmevans9:
This adds subscriptions to the GraphQL setup. This should work, for now. I am not currently handling connection IDs in subscriptions. I should be following up with an update for that either tonight or tomorrow.


rmevans9:
I have added tracking of clients along with the data for their subscriptions. For better or worse I stored the different clients data in an array under the subscription. This could mean we send all clients data down when we really only need one but I am envisioning the possibility of doing cross connection compares on state trees. Either way we can refactor later if needed

Issue #688, opened 2018-06-08
Reactotron logo too bright when not active tab


skellock:
Let's darken the inactive state of the first tab.

Its not entirely clear that the 2nd tab is the selected one here:

<img width="114" alt="image" src="https://user-images.githubusercontent.com/68273/41164359-01ad1c68-6b09-11e8-8b5c-d179b2103395.png">

rmevans9:
This is resolved.

PR #861, opened 2018-10-12
Lower brightness of Home-Button Logo

xstable:
Lower Brightness if not active.
Depend on Issue: https://github.com/infinitered/reactotron/issues/688


rmevans9:
LGTM! Thanks!

PR #871, opened 2018-10-28
Command clearing

rmevans9:
This still needs some work when we get the connection selecting setup done but this works for now to show how to do it.


mlaco:
Thanks for your PR, @rmevans9! Could you add a little more detail about what it does please?

PR #886, opened 2018-11-30
Added overlay to typings

lvlrSajjad:
used in andross boilerplate in Containers/App
```
export default DebugConfig.useReactotron
  ? console.tron.overlay(App)
  : App
```

if we convert app.js to app.ts , the error comes up so we need add overlay() to typings


rmevans9:
Thank you!

PR #889, opened 2018-12-02
Remove reactotron-redux from here

rmevans9:


rmevans9:
Merging this so we can continue our pursuit of cleaning everything up.

PR #890, opened 2018-12-03
Add support for wildcard selector

thibmaek:


rmevans9:
:wave: I am thinking that we might want to solve this problem closer to the UI. The reason for that is there are multiple state plugins (`redux` and `mobx-state-tree` and more coming hopefully) and I would want all of them to benefit from this change. I am thinking that a better spot to solve this would be here: https://github.com/infinitered/reactotron/blob/master/packages/reactotron-core-server/src/reactotron-core-server.ts#L290
What you might do is check path for a `*` and just ship in a empty string instead. This way all state plugins work the same.

rmevans9:
Hey @thibmaek - it turns out I have moved a lot of things around and this PR currently targets a file that was moved to a different repo. With that said I still think this can be a concern closer to the core. As such the new place that this code would need to be changed is here: https://github.com/infinitered/reactotron-core-server/blob/master/src/reactotron-core-server.ts#L290

If you don't have the time to implement the solution over there that is completely understandable and I will get it taken care of. Thanks!

thibmaek:
Alright, I'll try to do that tomorrow noon!

thibmaek:
@rmevans9 I created a new PR in the core package for this

PR #901, opened 2018-12-14
add tooltip to timeline and state header items

makenova:
When the `Reverse Order` and `Download` icons were added to the timeline header I didn't know what the did so I hovered over them hoping to see a tooltip.

This adds a tooltip to the header items for timeline and state like in the `CommandToolbar`.


rmevans9:
This is looking good. Only request is to also add the tooltip to the state backup tab. That icon is located here: https://github.com/infinitered/reactotron/blob/master/src/State/State.js#L72

src/State/State.js
```diff
@@ -69,10 +76,12 @@ class State extends Component {
           icon="import-export"
           renderActions={() => (
             <div>
-              <IconAddBackup
+              <Button
+                icon="file-download" 
+                onClick={stateBackupStore.sendBackup()}
```
rmevans9: This will execute the function at render time. We need to drop the `()` so we pass the function instead

rmevans9:
I am going to test this out tomorrow. If all goes well I will merge then. Thank you!

PR #921, opened 2019-02-14
Update core server to get the keepalive

### LINKED ISSUE #916: Disconnecting after a couple minutes
I'm using an external device (iPad Mini 2) to build my application on, I'm using Reactotron but I'm noticing after maybe 2-3 minutes of inactivity, the device will disconnect so I have to reload the app so it can re-make the call to Reactotron to re-hook. I'm developing a front-end system and a back-end system, so when I'm writing code for my back-end system It's kind of annoying have to reload my device because I've taken some time re-working some code. 

Why isn't the connection constant until I manually disconnect?

rmevans9:
Fixes #916

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "react-modal": "^3.4.4",
     "react-tap-event-plugin": "^3.0.2",
     "react-tooltip": "^3.5.1",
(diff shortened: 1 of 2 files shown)


rmevans9:
Holding on merge for a few min while I wrap up testing locally.

Issue #579, opened 2017-12-29
Save subcription keys between Reactotron launches


artem-russkikh:
It would be useful if I could save subscription paths between launches of Reactotron

skellock:
It’s a good idea for sure.

One of the things we will have to work out is how we keep multiple apps separate.

I wouldn’t want to see App 1’s settings while working on App 2.

skellock:
Tracking in #421

artem-russkikh:
@skellock note that "State snapshots" not equals "Subcriptions paths"

skellock:
Oh noes... you're right. Sorry!  🙇

PR #927, opened 2019-02-19
feat: Restore subscriptions upon app load

rmevans9:
fixes: #579

This makes it work by default. Might be worth handling this in a configuration at some point in the future when we have a good path to configuration values.


rmevans9:
Prior to merge I am sharing this feature out a little bit to see if we get any push back as it could change some users flow a little bit.

PR #1118, opened 2019-08-06
Add reapply overlay button

chrissloey:
Adds a "Reapply Overlay" button to the overlay page. My use case is that I often refresh the simulator (rather than hot reload), which then loses the overlay.

The button shows when there's an image uri. I guess ideally it would only show when the overlay isn't visible on the simulator anymore but I wasn't sure how to go about that.

I also updated the margin inputs to be `type="number"` so that arrow keys are usable in those fields.

Screenshot 2019-08-06 at 12 00 05

### DIFF
--- a/src/Native/NativeOverlay.js
+++ b/src/Native/NativeOverlay.js
@@ -31,12 +31,9 @@ const Styles = {
     marginTop: 20,
   },
   row: {
-    ...AppStyles.Layout.hbox,
-    padding: "15px 20px",
-    justifyContent: "space-between",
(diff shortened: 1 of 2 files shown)


chrissloey:
Prettier also made some style changes in NativeOverlayMargins.js, but it seemed to line up with the existing style in NativeOverlay.js so I just left the changes in.

rmevans9:
Hey @chrissloey! Thanks for the PR. I will be taking a closer look at this tonight and should hopefully be able to land it if there isn't any issues.

PR #1181, opened 2020-03-24
feat: deep search

maxkomarychev:
Currently reactoron only tests field if it is a string but this does not
work well if `tron.log` is called with arrays or objects.
    
This change tries to stringify value and test it against search
condition.

### DIFF
--- a/src/Stores/SessionStore.js
+++ b/src/Stores/SessionStore.js
@@ -69,8 +69,18 @@ class Session {
     }
 
     // safely matching the search term regexp against something passed in
-    const matching = value =>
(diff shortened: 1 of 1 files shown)


maxkomarychev:
well maybe that's a bit brute force solution. I could change it to traverse structures deeply but imo stringify is better since it's actually filtering by keys in objects
lmk if that's good, thanks :)

markrickert:
Thanks for this idea. This is probably a long time coming... closing this in favor of #1417. I'm going to have some people test it out and see if this adversely affects performance or not. And if not, we'll merge it. This should make timeline searching much better!

PR #1193, opened 2020-06-23
Adds homebrew instructions for easier install on mac.

markrickert:


rmevans9:
The homebrew version of reactotron is wildly out of date because of some mac signing issues I have to solve. I would hold off on doing this until we get to a point where we can effectively ship to Homebrew again.

markrickert:
@rmevans9 I tried installing it from the releases tab on github and it wouldn't open on my machine. When installing with homebrew it works great.

Not sure why there's a disconnect there 🤔

PR #1206, opened 2020-10-13
Added comments to Connections.js

souravnitkkr:


markrickert:
Thanks for the PR, but I don't think that this adds significant value to the codebase at the moment. Closing.

PR #1224, opened 2021-05-04
chore(docs): fix broken link

thebergamo:


morganick:
This example application no longer exists. We've removed this from the documentation and will add a new link in if we decide to add an example application back in.

PR #1227, opened 2021-05-22
Update plugin-redux.md

aCanalez:
Included a brief detail as to how to configure reactotron if using the configureStore function instead of the createStore function


kvlknctk (approved):
++1 awesome #1254

markrickert:
Thanks for this PR. It seems this has already been taken care of by #1405

PR #1230, opened 2021-06-21
Use the correct yarn command for adding to devDependencies

colbygarland:
The docs just had the incorrect flag for adding to dev dependencies,


markrickert:
Thanks for flagging this issue and for sending over a fix. Our docs have changed significantly since you opened this PR so I reimplemented your fix in #1415.

PR #1242, opened 2021-11-16
Add note to reload app after forwarding 9090

lindboe:
Small docs change to note that the app will need to be reloaded to show up in reactotron after forwarding port 9090.


mazenchami (approved):
👍

PR #1245, opened 2022-01-22
reactotronReactNative.default.createSagaMonitor is not a function

acevedo93:
Because createSagaMonitor does not appear to be a function, remove the () from the value and it works again.


markrickert:
I appreciate this contribution but Reactotron no longer supports redux sagas. We recommend using modern redux toolkit for redux development.

PR #1265, opened 2023-02-10
Migrate to yarn 3 for alpha

joshuayoes:
`workspace:*` does not work as expected in yarn 1. It will grab the most recent package off npm, not the latest version locally.

This PR updates the yarn 3 to allow for the user of `workspace:*`


lib/reactotron-core-ui/src/components/ActionButton/index.tsx
```diff
@@ -17,7 +17,7 @@ function ActionButton({ icon: Icon, tip, onClick }: Props) {
   return (
     <Container data-tip={tip} onClick={onClick}>
       <Icon size={24} />
-      <ReactTooltip place="bottom" />
+      <Tooltip place="bottom" />
```
joshuayoes, line 20: This seems to be causing the error during `yarn workspace reactotron-app run start` 
Screenshot 2023-02-09 at 6 17 36 PM

lib/reactotron-core-server/src/reactotron-core-server.ts
```diff
@@ -13,8 +13,7 @@ import {
   WssServerOptions,
 } from "./types"
 import { readFileSync } from "fs"
-import type { MittStatic } from "mitt"
-const Mitt: MittStatic = require("mitt") // Jest breaks with esm import, use cjs here ¯\_(ツ)_/¯
```
joshuayoes, line 17: Jest is not happy about handling for this for some reason :( https://app.circleci.com/pipelines/github/infinitered/reactotron/1388/workflows/753cdc32-7102-4549-8bc1-f65dd1de26c0/jobs/1565

PR #1268, opened 2023-02-28
Image overlay

carlinisaacson:
- Adds the image overlay feature back to the reactotron app
- Also adds a basic debug info view that will only show in development
  - This will allow developers to monitor and confirm the info coming into the react native view from the reactotron app
  - I did test this with a production build and it did not show
  - I also tested with an ignite app in production mode
    - Since we configure reactotron to do nothing in production the app didn't even connect to the reactotron app
- This was a lot of copying functionality from v2 and tweaking it to fit into v3
  - The major things that are different are react context vs mst and typescript
  - The types are odd because we are passing RN types around, but only reactotron-react-native has knowledge of RN types

iOS:

https://user-images.githubusercontent.com/1761434/222007202-17df4f41-28af-40af-a9d9-a90497e45af2.mov

Android:

https://user-images.githubusercontent.com/1761434/222007723-e6bd2b7f-5326-43eb-a0bf-b12150076098.mov


joshuayoes (approved):
Pushed up a small commit to improve the `.overlay()` type, but this works on a client codebase!

PR #1269, opened 2023-03-01
Map Command+K to clear timeline

carlinisaacson:
- Adds a key mapping for `command+k` to clear the timeline
- Also did a quick rearrange of the help screen so it makes a little more sense

I was too lazy to figure out a way to show key presses in the video. So you'll just have to trust that I'm pressing `command+k`.

https://user-images.githubusercontent.com/1761434/222011823-b481eceb-168b-4240-ae2f-59893ee40fc9.mov


joshuayoes (approved):
Ran this locally, works for me!

jamonholmgren (approved):

PR #1272, opened 2023-03-02
deps(*): Upgrade package dependencies to latest and greatest

silasjmatson:


joshuayoes (approved):
This looks good! Glad to see that all the important packages are not on the most recent major version (aside from ones we wanted to ignore like React, Rollup, styled-components)

I pushed up some commits:
- removed any reference to semantic release to avoid confusion
- fixed weird yarn.lock issue where workspace packages were resolving to their latest npm version, not the local workspace (this may fix Carlin's issue in the last release)
- added a script to tests local changes in another app, like ignite
- added some missing typecheck scripts

This is good to merge on Monday. Deploying late day on Friday is bad juju

PR #1275, opened 2023-03-06
fix(release): Pin electron correctly, allow building unsigned release builds locally

silasjmatson:


apps/reactotron-app/scripts/build.release.js
```diff
@@ -1,29 +1,41 @@
...
-}
-if (BUILD_TARGET === "macos") {
-  console.log(`MacOS Code Signing Certificate found at: '${CSC_LINK}'`)
+if (isCi) {
+  const path = require("path")
+  const fs = require("fs")
+  const CSC_LINK = path.join(
+    __dirname, // /apps/app/scripts
+    "..", // /apps/app/
+    "Certificates.p12" // /apps/app/Certificates.p12
```
joshuayoes, line 17: Non-blocking, but would be nice:
```suggestion
  const CSC_LINK = path.join(
    __dirname, // ~/apps/reactotron-app/scripts
    "..", // ~/apps/reactotron-app/
    "Certificates.p12" // ~/apps/reactotron-app/Certificates.p12
```

apps/reactotron-app/scripts/build.release.js
```diff
@@ -1,29 +1,41 @@
...
+  )
+  if (BUILD_TARGET === "macos" && !fs.existsSync(CSC_LINK)) {
+    throw new Error(`CSC_LINK not found at ${CSC_LINK} for ${BUILD_TARGET} target}`)
+  }
+  if (BUILD_TARGET === "macos") {
+    console.log(`MacOS Code Signing Certificate found at: '${CSC_LINK}'`)
+  }
+} else {
+  console.log("Not running in CI, skipping code signing")
+  console.log("If you want a signed build, run electron-builder directly")
```
joshuayoes, line 27: Non-blocking, but would be nice:

```suggestion
  console.log("For production builds: electron-builder needs a MacOS Developer Certificate in the MacOS Keychain or linked at CNC_LINK variable. See: https://www.electron.build/code-signing.html")
```

joshuayoes (approved):
Added a few nice to haves, but looks good

PR #1284, opened 2023-03-22
fix(reactotron-react-js): Add project.json so reactotron-react-js will release properly

silasjmatson:


joshuayoes (approved):
As long as we have a git tag for reactotron-react-js@2.1.1 pushed up to the repo, this is good to merge!

PR #1288, opened 2023-04-20
Add "react-native" package.json entry

joshuayoes:
This PR adds `react-native` entry key and `src` to the files included in the published packages.

This will allow Metro to handle bundling the Typescript source files directly, which can allows Metro to offer better errors and include source maps.

Adding this to your metro config will tell Metro to prefer this entry point instead of `main`. 

Screenshot 2023-04-20 at 4 38 38 PM

Example:

| Before | After |
|--------|--------|
| Screenshot 2023-04-20 at 4 31 47 PM|Screenshot 2023-04-20 at 4 33 43 PM |

### DIFF
--- a/lib/reactotron-apisauce/package.json
+++ b/lib/reactotron-apisauce/package.json
@@ -12,10 +12,12 @@
   "files": [
     "dist",
     "LICENSE",
-    "README.md"
+    "README.md",
+    "src"
   ],
   "main": "dist/index.js",
   "typings": "dist/types/src/index.d.ts",
+  "react-native": "src/index.ts",
   "scripts": {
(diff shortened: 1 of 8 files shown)


kevinvangelder (approved):
Would it also be a good idea to document the required metro.config changes in the readme?

joshuayoes:
> Would it also be a good idea to document the required metro.config changes in the readme?

Yeah that would be a good idea. I also was thinking it would be good to include this in Ignite by default once the beta is launched

joshuayoes:
Resolved in favor of https://github.com/infinitered/reactotron/pull/1326

PR #1291, opened 2023-06-14
feat(lib): add `react-native-mmkv` plugin

mazenchami:
Adding a [`react-native-plugin`](https://github.com/mrousavy/react-native-mmkv) for `Reactotron`


lib/reactotron-react-native-mmkv/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+  "compilerOptions": {
+    "allowJs": false,
+    "declaration": true,
+    "declarationDir": "dist/types",
+    "rootDir": "src",
+    "emitDeclarationOnly": true,
+    "emitDecoratorMetadata": true,
+    "allowSyntheticDefaultImports": true,
+    "experimentalDecorators": true,
+    "module": "es2015",
```
ryanlntn, line 12: Any reason not to just go `esNext`?

ryanlntn (approved):

lib/reactotron-react-native-mmkv/tsconfig.json
joshuayoes: This was copy and pasted over from `reactotron-mmkv`.

There is a lot of legacy `tsconfig.json` settings and rollup stuff that could be simplified, but I think for now it's worth keeping the same until we have a cohesive strategy across all the modules

joshuayoes: Eventually, it would be nice to have all of these extend from `@tsconfig/react-native/tsconfig.json` or something like that and drop rollup.

joshuayoes (approved):
This looks good! (Especially since we wrote it on a call together 😆)

I'll merge this tomorrow when I have some more time to stand by for any issues in the CI, since this is a new package.

PR #1295, opened 2023-07-18
Make plugin clients more generic, add types for state responses

joshuayoes:
This PR makes changes to address this issue in Ignite: https://github.com/infinitered/ignite/pull/2446#issuecomment-1636506481

This PR resolves the need for `// @ts-expect-error` by:
- adding a generic to the `reactotron-mst` plugin types
- adding types to the payloads for the state messages for `reactotron-mst` and `reactotron-redux`
- adds missing `reactotron-react-native` commands to contracts

### DIFF
--- a/lib/reactotron-mst/src/reactotron-mst.ts
+++ b/lib/reactotron-mst/src/reactotron-mst.ts
@@ -53,7 +53,7 @@ import {
 
 // --- Helpers ---------------------------------
 
-const dotPath = (fullPath: string, o: any) => path(split(".", fullPath), o)
+const dotPath = (fullPath: string, o: Record<string, any>) => path(split(".", fullPath), o)
 const isNilOrEmpty = (value: any) => isNil(value) || isEmpty(value)
 const isReactNativeEvent = (value: any) =>
   typeof value !== "undefined" &&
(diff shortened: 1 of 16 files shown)


joshuayoes:
Tests are failing due to some runtime changes. I've got a stashed commit with the fix for some of them, I'll come back to this

lib/reactotron-mst/src/reactotron-mst.ts
```diff
@@ -263,10 +284,15 @@ export function mst(opts: MstPluginOptions = {}) {
      *
      * @param command A reactotron command.
      */
-    function backup(command: any) {
-      const trackedNode = trackedNodes[command.mstNodeName || "default"]
+    function backup(command: Command<"state.backup.request">) {
+      const trackedNode =
+        trackedNodes[
+          "mstNodeName" in command && typeof command.mstNodeName === "string"
```
joshuayoes, line 290: No client sends a command with this key. So this boolean always evaluates to 'default'. Future TODO would be to figure out what the actually desired behavior is here

joshuayoes:
Installed these package versions into ignite and they seem happy now.

PR #1297, opened 2023-07-24
Fix eslint setup

joshuayoes:


joshuayoes:
Master now already has more recent eslint upgrades, closing this

PR #1309, opened 2023-09-28
fix(reactotron-app) Dev tools now opens automatically every time the app launches in dev mode

markrickert:
The code was already in there to open the dev tools when the app was launched, but it wasn't happening because of electron's startup procedure. 

I moved the call to `window.webContents.openDevTools()` inside the `ready-to-show` lifecycle method and now in development mode, the reactotron app will automatically show the dev console.


joshuayoes (approved):
Tested this locally, nice change!

PR #1313, opened 2023-10-13
chore(dependencies): Upgrade lots of packages.

markrickert:
This PR modernizes lots of old dependencies and aligns the dependencies between monorepo packages which were previously disparate packages.

Of particular note, the upgrade of `styled-components` and subsequent syntax changes got rid of the warnings and errors in the reactotron-app's console regarding invalid attributes...

These errors and warnings have been squashed in this branch:

Screenshot 2023-10-12 at 9 40 52 PM


joshuayoes (approved):
Ran and tested this locally, it loads up super fast! Thank you for taking us to the modern age.

I pushed up a commit to fix a nit of changing this type annotation to `React.FC<React.ReactNode>` to `React.FC<{ children: React.ReactNode }>`. Technically this passes typechecking but we to pass down the children prop for the ReactNode, not the whole props object as a ReactNode.

Great work on this Mark!

PR #1315, opened 2023-10-16
chore(docs): Move around and organize docs for impending `ir-docs` integration

markrickert:
This PR reorganizes the docs folder to better fit within the framework for the upcoming docs integration with https://github.com/infinitered/ir-docs

It adds metadata, `_category_.json` file for docusauras, and some extra content for index files.


trevor-coleman (commented):
Remove this from here -- leave it alone elsewhere.

.circleci/config.yml
```diff
@@ -9,6 +9,7 @@ defaults: &defaults
 version: 2.1
 orbs:
   node: circleci/node@5.1.0
+  publish-docs: infinitered/publish-docs@0.4.8
```
frankcalise, line 12: I think this can be bumped to 0.4.9

https://circleci.com/developer/orbs/orb/infinitered/publish-docs

.circleci/config.yml
```diff
@@ -294,6 +295,28 @@ workflows:
                 - beta
                 - alpha
 
+  publish_to_docs_site:
+    jobs:
+      - publish-docs/publish_docs:
```
frankcalise, line 300: There is also a build_docs workflow on the documentation site, is that needed here as well?

.circleci/config.yml
markrickert: Done!

.circleci/config.yml
markrickert: That might be a question for @trevor-coleman. I'm unsure if this CI script is set up correctly. This branch was made a few months ago and things might have changed with the CI setup.

trevor-coleman: build-docs is "only build" -- good for automated testing so PRs don't get merged with broken links etc.
publish-docs is "build and publish" -- so it builds it, and pushes it to `ir-docs`

trevor-coleman:
> Still want to pull it down and try it but I think we can bump the orb version from the looks of the docs

@frankcalise  -- I set the version to '0.4' which works like `^0.4.0` in a `package.json`

I have them all set that way, so that I can make non-breaking changes without having to update them all.

trevor-coleman:
Well that's working at least.

CleanShot 2023-12-13 at 14 38 44@2x


Going to try the force-publish

PR #1318, opened 2023-10-20
fix(content-view): parse stringified object

### LINKED ISSUE #1317: Convert request tab into object tree instead of string
When you inspect an API request, it is just stringified, compared to when you inspect the response you can expand/collapse the object tree

Repo steps:
1. Open reactotron
2. Load up an app
3. Make a network request
4. Go to request tab and see that it is stringified
5. See response tab
6. See that it is an object tree

frankcalise:
## Why
- Closes #1317

Currently, if a request/response body comes back as a stringified JSON object (which at the very least, Apollo Client is doing for request bodies), the `TimelineCommand` displays it as a string.

However, there are times where that string is a stringified object and we could do better to give it the tree view we are used to in Reactotron.

## How

Check if the incoming value prop can be handled with `JSON.parse`, if so then display it as an object. Otherwise let all the other string and fallback formatters happen.

### Before

image

### After

image

### DIFF
--- a/lib/reactotron-core-ui/src/components/ContentView/index.tsx
+++ b/lib/reactotron-core-ui/src/components/ContentView/index.tsx
@@ -27,7 +27,14 @@ export default function ContentView({ value, treeLevel }: Props) {
   if (value === null) return <NullContainer>null</NullContainer>
   if (value === undefined) return <UndefinedContainer>undefined</UndefinedContainer>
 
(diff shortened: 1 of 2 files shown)


joshuayoes (approved):
Gave this a test, nice work!

https://github.com/infinitered/reactotron/assets/37849890/92e6dacb-30c5-4874-854c-00eded564d06

PR #1319, opened 2023-10-24
fix(display-cmd): correct preview text

frankcalise:
Fixes the Display command showing the `title` prop in the `preview` portion of the Display command in the Timeline


joshuayoes (approved):
This has bugged me for a while, thanks for fixing this 🔥

PR #1311, opened 2023-09-29
fix(reactotron-react-native): remove rn-host-detect, use embedded snippet

joshuayoes:
There are several issues where users had trouble connecting to reactotron #1305, #1107, #815

This PR fixes them by upstreaming the fix into `reactotron-react-native` itself.

How to test:
1. `yarn`
2. `npx ignite-cli new PizzaApp --yes`
3. `npx zx scripts/install-workspace-packages-in-target.mjs ../PizzaApp`
4. `cd PizzaApp && yarn expo:android`
7. See that reactotron connects (without using `adb reverse tcp:9090 tcp:9090`)

Issue #1107, opened 2019-06-21
Can't get Reactotron to Connect to Expo App


cdcv:
Title says it all.
When I run basic test app on iOS simulator, I get 'Reactotron Configured' in the console log, but Reactotron app does not connect on Mac.
Have tried *a lot* of configuration variants etc based on prior posts.
Currently, trying to do something very basic... just get a bare expo react native app to run on a simulator and connect to Reactotron.
I've downloaded the latest version, and followed the react native install instructions (many times).

Currently using the simplest config options:
in ReactotronConfig.js:
import Reactotron from 'reactotron-react-native'
Reactotron
    .configure({
        name: "React Native Demo",
    }) // controls connection & communication settings
   // .useReactNative() // add all built-in react native plugins
    .connect() // let's connect!

in App.js:
if(__DEV__) {
  import('./ReactotronConfig').then(() => console.log('Reactotron Configured'))
}

from package.json:
    "reactotron-react-native": "^3.6.4"
    "expo": "^33.0.0",
    "react": "16.8.3",

Thanks for your help.

nzrubin:
same. I'm on windows running expo using an android simulator. How many use expo? what is the point of reaction if it cannot run on Expo

GollyJer:
When using Expo you have to set the host url to your local machine's ip address.

```typescript
// ReactotronConfig.js:
import Reactotron from 'reactotron-react-native'

Reactotron
    // Your real ip address 👇
  .configure({ host: '192.168.0.100' })
  .useReactNative() 
  .connect()
```

jaho-design:
Same problem here with lastest expo version. 
I try to configure host but didn't work

// package.json
  "dependencies": {
    "expo": "^34.0.1",
    "react": "16.8.3",
    "react-dom": "^16.8.6",
    "react-native": "https://github.com/expo/react-native/archive/sdk-34.0.0.tar.gz",
    "react-native-web": "^0.11.4"
  },
  "devDependencies": {
    "babel-preset-expo": "^6.0.0",
    "reactotron-react-native": "^3.6.4"
  },

pkail:
On Mint 19.1/Expo 33
 My app loads and runs correctly but Reactotron doesn't react, either with or without host IP as suggested above.  Such a pity!  It looked good in the presentation.

GollyJer:
@jaho-design @pkail 

[...]

PR #1320, opened 2023-11-01
feat(reactotron): Add Android `adb` command interface to the Help page.

markrickert:
Sometimes connecting to physical android devices can trip up people new to reactotron or react native development.

This PR adds a section to the Help page where users can see which android devices that are connected to their computer via USB and automatically run common functions on each device including:

1. `adb reverse tcp:9090 tcp:9090` for reactotron connections on android devices
2. `adb reverse tcp:8081 tcp:8081` for metro connections on android devices
3. `adb shell input text '"RR"` to reload an active react native application
4. `adb shell input keyevent 82` to shake the device and display the react native developer menu. This is particularly useful for simulator instances

...all with a simple to use interface on the Help page. This works with multiple connected devices and will only run on the selected device. 

There's also a hidden "port settings" for the tcp reverse command that you can access via the "gear" settings icon where power users can change the port that reactotron and metro are running on.

### To Do:

- [x] Test that this works on Windows instances
- [ ] Fail gracefully if `adb` isn't set up correctly
- [ ] Possible: feedback to the user that something actually happened when they click "Reverse Tunnel"
- [ ] Determine a better place for this, maybe? Should this go on the Help page?

### How It works:

When the app boots up, it defines the actions that can be taken with `adb` as well as spawns a process for `adb track-devices` so that we can be notified whenever an android device is connected or disconnected from USB.

When the Help screen loads, it fires off a command to get the list of connected devices and parses the result from the command's `stdout`. If the `adb track-devices` process determines that a device has been connected or disconnected, it requests the list of devices again.

Overall, this works REALLY well!

### Here's the functionality in action:

> I am physically plugging and unplugging my android device:

667EE0DF-9FE5-41F4-BA0A-21C23C4F2091-9010-000064220203A491

Here's what the interface looks like with multiple android devices connected:

Screenshot 2023-11-01 at 5 41 10 PM

Here's the input fields where power users can change the reactotron and metro ports:

Screenshot 2023-11-01 at 5 46 20 PM

### DIFF
--- a/apps/reactotron-app/src/main/index.ts
+++ b/apps/reactotron-app/src/main/index.ts
@@ -6,6 +6,7 @@ import { autoUpdater } from "electron-updater"
 import windowStateKeeper from "electron-window-state"
 
 import createMenu from "./menu"
+import { setupAndroidDeviceIPCCommands } from "./utils"
 
 const isDevelopment = process.env.NODE_ENV !== "production"
 
@@ -32,8 +33,8 @@ function createMainWindow() {
     width: mainWindowState.width,
     height: mainWindowState.height,
     titleBarStyle: "hiddenInset",
-    webPreferences: { 
-      nodeIntegration: true, 
+    webPreferences: {
+      nodeIntegration: true,
       contextIsolation: false,
       webgl: false, // Disable webGL for performance reasons
       spellcheck: false, // Disable spellcheck for performance reasons
@@ -47,7 +48,7 @@ function createMainWindow() {
 
     if (isDevelopment) {
       window.webContents.openDevTools()
-    }  
+    }
   })
 
   window.setBackgroundColor("#1e1e1e") // see reactotron-core-ui for background color
@@ -97,4 +98,7 @@ app.on("activate", () => {
 // create main BrowserWindow when electron is ready
 app.on("ready", () => {
   mainWindow = createMainWindow()
+
+  // Sets up the electron IPC commands for android functionality on the Help screen.
+  setupAndroidDeviceIPCCommands(mainWindow)
 })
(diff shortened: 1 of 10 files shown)


frankcalise:
Tested Windows and Ignite v9, running Expo Go with great success! 🎉 

With `adb` set up correctly, it found my device wired to my machine properly. Disconnecting drops the connection in the Help tab.

All 3 buttons worked as well
- Reload app and shake worked right away
- Reverse tunnel worked as well, timeline showed up with boilerplate after reloading app

## Screenshot

image

frankcalise (approved):
The tooltips are a great add, I think we should style them better so they're more visible. They blend a little too much.

image

I think if we outlined them more like the code snippet formatting they would be easier to read, something like this: 

image

As far as the location of the Android devices connected via USB, maybe it could live above help as `Devices`, but then we'd maybe expect to see iOS ones there? Feels like it could exist on the connections bottom bar in some fashion, where it could say `0 connections, 1 device` (before your ran reverse tunnel / reload, for example) and maybe initiate the commands from there some how?

I think @jamonholmgren had ideas on reworking that bottom bar into a tab possibly, anyway as it's hard to discover.

markrickert:
I've updated this branch with a new default tooltip style that includes a border and text color based on the app's theme.

Screenshot 2023-11-02 at 12 08 32 PM

markrickert:
I'm wondering if this might be able to be integrated into the "Home" screen for more discoverability? The home screen title is "Connections" and it shows all the apps that are connected to reactorton.

This might take some design rework though.

 Screenshot 2023-11-02 at 12 31 41 PM

joshuayoes (approved):
This is a great addition Mark! I played around with this and it is very helpful!

It helped me realize that https://github.com/infinitered/reactotron/issues/1107 is not actually resolved... I thought I resolved this with https://github.com/infinitered/reactotron/pull/1311 but being able to run `adb reverse` easily from the UI would be a better solution for now.

I'm not sure that the help page is the best place for this for now, but I don't have an idea of where it would be better, so I think revisiting it in a future would be a good idea.

PR #1322, opened 2023-11-06
chore(tests): Migrate tests to use `get-port-please` and update some dependencies

markrickert:
I tried updating our devDependency `get-port` from `5.1.11` to the latest `7.0.0` and ran into a bunch of problems with the tests. I found an alternative package that works called [`get-port-please`](https://github.com/unjs/get-port-please) that worked flawlessly without any additional babel or jest configuration changes.

I also updated `styled-components`, `testdouble`, and various `eslint` packages across the board.


joshuayoes (approved):
Nice work here Mark!

PR #1327, opened 2023-12-01
feat(reactotron-react-native): enable trackGlobalLogs in useReactNative by default

joshuayoes:
This PR adds a new core plugin called `trackGlobalLogs` to the `useReactNative` default plugins, which will automatically track `console.log`, `console.warn`, and `console.debug` without needing to do calls to `Reactotron.log` or `console.tron.log`.

This has been enabled in Ignite for [several releases now](https://github.com/infinitered/ignite/blob/v9.0.4/boilerplate/app/devtools/ReactotronConfig.ts#L149-L178), so we will be able to remove this snippet. 

Monkey patching globals like console log is not ideal but we determined that we would expect a debugging tool to collect logs. We also already do similar strategies for inspecting network traffic, so we think this is a tradeoff worth making.


joshuayoes:
This will requite a new PR to ignite to update to the latest version [to remove this snippet](https://github.com/infinitered/ignite/blob/v9.0.4/boilerplate/app/devtools/ReactotronConfig.ts#L149-L178), and remove some references to `console.tron.log` to prevent duplicate logs

PR #1335, opened 2023-12-07
fix: Display error if port is already in use

### LINKED ISSUE #1334: Reactotron fails silently when port 9090 is already in use
## Problem

I tried running a Reactotron 2.8.1 instance and a Reactotron 3.0.0-beta.39 instance next to each other.

Two Reactotron instances

Whichever order I open them in, the first one "wins", and the second one just fails silently.

So, to test this conflict, I ran this one-liner to start a new websocket server:

```bash
node -e "const http = require('http'); const WebSocket = require('ws'); const server = http.createServer((req, res) => { res.end('HTTP Server Running'); }); const wss = new WebSocket.Server({ server }); wss.on('connection', ws => { ws.on('message', message => { console.log('received: %s', message); }); console.log('WebSocket client connected'); }); server.listen(9090); console.log('HTTP and WebSocket server running on port 9090');"
```

Sure enough, starting Reactotron 2 or 3 up resulted in them failing silently. Instead, the websocket server would log all of the Reactotron communication attempts:

websocket server

## Proposed Solution

I think we should detect when the websocket server fails to start due to a port conflict and display a UI that says "Port 9090 appears to already be in use. [Retry button]".

jamonholmgren:
Fixes #1334.

This adds an icon to the left sidebar that indicates the server status: Stopped, Started, or Port Unavailable.

<table>
<tr><th>Server Started</th><th>Port 9090 Unavailable</th></tr>
<tr><td>

</td>
<td>

</td>
</tr>
</table>

Clicking the icon reloads Reactotron, which is a quick & dirty way to attempt reconnection. In the future, we can handle it more elegantly.

### DIFF
--- a/apps/reactotron-app/src/renderer/components/SideBar/index.tsx
+++ b/apps/reactotron-app/src/renderer/components/SideBar/index.tsx
@@ -1,12 +1,14 @@
 import React, { useContext } from "react"
 import LayoutContext from "../../contexts/Layout"
+import StandaloneContext from "../../contexts/Standalone"
 
-import SidebarStateless from "./Stateless"
+import SidebarStateless from "./Sidebar"
 
 function SideBar() {
   const { isSideBarOpen } = useContext(LayoutContext)
+  const { serverStatus } = useContext(StandaloneContext)
 
-  return <SidebarStateless isOpen={isSideBarOpen} />
+  return <SidebarStateless isOpen={isSideBarOpen} serverStatus={serverStatus} />
 }
 
 export default SideBar
--- a/apps/reactotron-app/src/renderer/contexts/Standalone/useStandalone.ts
+++ b/apps/reactotron-app/src/renderer/contexts/Standalone/useStandalone.ts
@@ -2,13 +2,18 @@ import { useCallback, useReducer } from "react"
 import { produce } from "immer"
 
 export enum ActionTypes {
+  ServerStarted = "SERVER_STARTED",
+  ServerStopped = "SERVER_STOPPED",
   AddConnection = "ADD_CONNECTION",
   RemoveConnection = "REMOVE_CONNECTION",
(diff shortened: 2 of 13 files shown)


apps/reactotron-app/src/renderer/components/SideBar/index.tsx
```diff
@@ -1,12 +1,14 @@
 import React, { useContext } from "react"
 import LayoutContext from "../../contexts/Layout"
+import StandaloneContext from "../../contexts/Standalone"
 
-import SidebarStateless from "./Stateless"
```
jamonholmgren, line 4: Not really a fan of these Stateless.tsx files.

apps/reactotron-app/src/renderer/contexts/Standalone/useStandalone.ts
```diff
@@ -30,13 +35,16 @@ export interface Connection extends ReactotronConnection {
...
+  serverStatus: ServerStatus
   connections: Connection[]
   selectedClientId: string
   orphanedCommands: any[] // Command[]
   commandListeners: ((command: any) => void)[] // ((command: Command) => void)[]
 }
 
 type Action =
+  | { type: ActionTypes.ServerStarted; payload: undefined }
+  | { type: ActionTypes.ServerStopped; payload: undefined }
```
jamonholmgren, line 47: I honestly think we should implement MobX-State-Tree in Reactotron. I am not a big fan of the useReducer & provider/context DX.

lib/reactotron-core-server/README.md
```diff
@@ -12,38 +12,32 @@ import { createServer } from "reactotron-core-server"
 // configure a server
 const server = createServer({
   port: 9090, // default
-
-  onStart: () => null, // fires when we start the server
```
jamonholmgren, line 16: These examples were wrong. I fixed them.

PR #1338, opened 2023-12-08
chore(reactotron-mst): remove test-double

joshuayoes:
This is in preparation for unlocking bun test in the future.

`testdouble` creates a cryptic package error when running in `bun test` on this workspace.

This PR refactors to use jest.fn and other mocking features instead of `testdouble`.

| Before | After |
|--------|--------|
| Screenshot 2023-12-08 at 2 32 59 PM | Screenshot 2023-12-08 at 2 33 47 PM |

Bun is 200x times faster than Jest 🤯 

2s isn't bad for Jest for these tests but I think adding them up across the repo will help speed up the CI a lot.


joshuayoes:
In the future, migrating this suite to bun will be a 1 liner change
Screenshot 2023-12-08 at 2 47 21 PM

PR #1376, opened 2023-12-18
ci: require manual approval before building docs from forks

trevor-coleman:
Separates out build step, so there is one for builds from forks that requires approval before building the docs.

This is prevent unreviewed code from being run in the CI environment.


trevor-coleman:
This is no longer needed as we've implemented the `ci:trust` script to run tests on forks.

https://github.com/infinitered/reactotron/pull/1408

PR #1380, opened 2023-12-19
WIP: Convert app state to MobX-State-Tree

jamonholmgren:
Currently in-progress moving reactotron-app to MST.


apps/reactotron-app/src/renderer/models/Connection.ts
```diff
@@ -9,7 +9,7 @@ export const ConnectionModel = t
 
     platform: t.enumeration(["ios", "android", "browser"] as const),
     name: "",
-    platformVersion: "",
+    platformVersion: t.optional(t.union(t.number, t.string), ""),
```
coolsoftwaretyler, line 12: @jamonholmgren - `platformVersion` is coming back as a number on Android, and a string on iOS. This handles both and defaults to an empty string.

PR #1381, opened 2023-12-19
Improve getting host name in example for react-native.md

jcubic:
Improve and simplify the example in the docs. To get the hostname it's better to use [URL constructor/API](https://developer.mozilla.org/en-US/docs/Web/API/URL_API).


jcubic:
You have broken CI/CD got error:

> fatal: Remote branch pull/1381 not found in upstream origin

jcubic:
I'll update the commit message, I'm cloning the repo. It's huge.

joshuayoes:
URL does not work in React Native by default
https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Blob/URL.js#L174

There are polyfills you can use, but apps do not have it by default.

However, as of [this PR](https://github.com/infinitered/reactotron/pull/1311), this snippet is no longer needed. We should remove this section from the readme.

jcubic:
If the URL doesn't work why don't you use regular expressions?

```javascript
'https://example.com/path/to/file'.match(/:\/\/([^\/:]+)/)[1]
'https://example.com:8080/path/to/file'.match(/:\/\/([^\/:]+)/)[1]
```

joshuayoes:
We could totally do that as well, this snippet has just been proven to work and is what's been floating around issues and docs for a few years.

jcubic:
Ok, I'm closing this as invalid.

PR #1386, opened 2023-12-20
Adds tooltips to state snapshot item commands.

markrickert:
Previously, the state snapshot buttons were hard to understand what they did (other than maybe the trash can icon)

This PR adds Tooltip items to each button so that users can more easily understand what the buttons are going to do.

Screenshot 2023-12-20 at 4 11 53 PM

Screenshot 2023-12-20 at 4 12 09 PM

Screenshot 2023-12-20 at 4 12 16 PM

Screenshot 2023-12-20 at 4 12 21 PM


frankcalise (approved):
love this addition!

Issue #1359, opened 2023-12-14
Feature: Jokes in the timeline


jamonholmgren:
When we have a blank timeline, it's a opportunity to add some jokes. Let's get 10+ debugging jokes and add a random one to the screen when the timeline is empty.

![CleanShot 2023-12-14 at 15 47 10@2x](https://github.com/infinitered/reactotron/assets/1479215/2acc3ea2-b194-46e1-805f-221c8290c10b)

We should also add a link to the Quick Start guide while we're at it.

https://docs.infinite.red/reactotron/quick-start/react-native/

markrickert:
@jamonholmgren Is this still necessary now that the home screen isn't the timeline screen? 

Related: https://github.com/infinitered/reactotron/pull/1393

jamonholmgren:
I think if the timeline is empty it would still be fun, but this isn't critical

PR #1395, opened 2024-01-05
Timeline empty state improvements

carlinisaacson:
https://github.com/infinitered/reactotron/issues/1359

- Timeline empty state improvements
  - Adds a button for the quick start guide
  - Adds a random developer joke

Contributors: @carlinisaacson @trevor-coleman @darinwilson 

DD321274-A94C-42AF-BCAF-AB9A6996B126


lib/reactotron-core-ui/src/components/RandomJoke/RandomJoke.tsx
```diff
@@ -0,0 +1,73 @@
...
+  },
+]
+
+export function RandomJoke() {
+  const jokeRef = useRef(0)
+
+  useEffect(() => {
+    const ref = Math.floor(Math.random() * jokes.length)
+    jokeRef.current = ref
+  })
```
robinheinze, line 65: Question: Is the dependency array left out here intentionally so a new joke will appear on every re-render?

robinheinze (approved):
Approved! Just left one question :)

lib/reactotron-core-ui/src/components/RandomJoke/RandomJoke.tsx
carlinisaacson: We wanted it to only run on mount. So we had an empty array. Then eslint was giving us a warning to either add the jokes.length dependency or remove the array. So we removed it. Then we extracted it to core-ui, and now when I add the empty array back I don't get the warning there. I personally would rather the empty array be there so I'll add it back in.

morganick (approved):
LGTM 👍 and works as expected.

Note: The screenshot on the docs needs to be updated: https://docs.infinite.red/reactotron/quick-start/react-native/

frankcalise (approved):
updated with latest from main

PR #1400, opened 2024-01-11
chore(deps): upgrade query-string to latest version

m00nreal:
Upgraded to latest version of query-string.

The breaking changes presented in version `7.0.0` does not affect the current usage in the repo.

By the way, have you considered removing the dependency? I searched across the repo and it is only used once in one file.

Anyway, I did some smoke testing and everything seems to be working fine (and ran the test locally, everything went smooth).

### DIFF
--- a/lib/reactotron-react-native/package.json
+++ b/lib/reactotron-react-native/package.json
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "mitt": "^3.0.1",
-    "query-string": "6.14.1",
+    "query-string": "8.1.0",
     "reactotron-core-client": "workspace:*"
   },
   "optionalDependencies": {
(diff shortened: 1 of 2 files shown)


markrickert:
[Version 8 of `query-string` moved to a pure ESM version](https://github.com/sindresorhus/query-string/releases/tag/v8.0.0) instead of cjs. Do we know how that may affect the monorepo @joshuayoes? 

I think I tried updating this package in the past and failed because of this. https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

I think if that's going to be a huge change, the most recent version we could use right now would be `7.1.3`

joshuayoes:
Yeah the ESM upgrade was an issue, I believe that's why I left it at the version it's at. 

This is the only place that we use it, so we could find another package or embed a snippet  https://github.com/infinitered/reactotron/blob/master/lib/reactotron-react-native/src/plugins/networking.ts#L76

PR #1401, opened 2024-01-11
fix(reactotron-app): added fontFamily from useTheme to Modal component

m00nreal:
The modals were not applying the font family from the theme. 
Small change but I think it looks much better.

**Before:**  
Timeline Filter
Add Subscription
Rename Snapshot
Dispatch Action

**After**
Timeline Filter
Add Subscription
Rename Snapshot
Dispatch Action


markrickert (approved):
This looks great! Thanks for the contribution!

m00nreal:
I was wondering why the pipeline for build_docs is failing?
Is that something I have to fix on my end?

markrickert:
@m00nreal I just merged it. There was an issue with our CI server, nothing on your end. Thanks again for the fix!

PR #1404, opened 2024-01-16
fix(reactotron-redux): Typescript fixes for reactotron-redux interface

Jpoliachik:
Adding `reactotron-redux` to a new Ignite app was causing Typescript issues for
```
.use(reactotronRedux())
```

I noticed that `reactotron-mst` worked fine, and the two had similar compiled TS definitions, except for the placement of generics. 

`reactotron-redux - index.d.ts`: 
```
declare function reactotronRedux<Client extends ReactotronCore = ReactotronCore>(pluginConfig?: PluginConfig): (reactotron: Client) => {

```
vs
`reactotron-mst - index.d.ts`:
```
export declare function mst(opts?: MstPluginOptions): <Client extends ReactotronCore = ReactotronCore>(reactotron: Client) => {
```

So as a result, I copied a bit of `reactotron-mst`'s plugin function to help influence the compiled output. 

I tested by copying the compiled output over to my Ignite project's `node_modules/reactotron-redux` and it worked 👍

### DIFF
--- a/lib/reactotron-redux/src/index.ts
+++ b/lib/reactotron-redux/src/index.ts
@@ -24,9 +24,13 @@ function reactotronRedux<Client extends ReactotronCore = ReactotronCore>(
     })
   }
 
-  return (reactotron: Client) => {
+  function plugin<Client extends Reactotron = Reactotron>(reactotron: Client) {
     return {
+      // Fires when we receive a command from the Reactotron app.
       onCommand: createCommandHander(reactotron, mergedPluginConfig, onReduxStoreCreation),
+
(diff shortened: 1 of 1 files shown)


markrickert (approved):
Nice! I noticed type issues last time I integrated a redux project with reactotron. 

Note that we have a helper script that you can run to automatically shim your demo project's `node_modules` with compiled versions of `lib` projects - https://docs.infinite.red/reactotron/contributing/#making-changes-to-other-monorepo-packages-that-live-in-lib

joshuayoes (approved):
Nice job here!

PR #1405, opened 2024-01-16
chore(docs): update reactotron-redux docs to include redux-toolkit

Jpoliachik:
It took some time for me to figure out how to properly integrate `reactotron-redux` with Redux Toolkit's `configureStore`. Since Redux Toolkit is the recommended method now, we should include in docs.


mazenchami (approved):
LGTM! 🎉 🚀

Issue #723, opened 2018-07-03
Analytics


skellock:
@kemiller asked this about a month ago.... do we know how many people use Reactotron?

My answer:  🤷‍♂️ 

Let's find out ... but let's do it in a non-douchey way where people don't think we're stealing their information.

Opt-in would be preferable, but nobody will turn it on.

Perhaps show them exactly what the payload looks like if people are curious?

skellock:
We can put the off switch in the upcoming #721 config file.

jamonholmgren:
I’d love to see this info, but agree it needs to be done in a non-douchy a way as possible without making it an option nobody turns on.

Perhaps there’s a way to make a compelling case for leaving it on in the message that tells people that we are doing this. I have some ideas; I’d love to show some examples of what I’m thinking once I get a few minutes to put together a mock-up?

skellock:
I'm all ears, as I have pretty much 0 ideas on how to do this right! haha.

jamonholmgren:
So this isn't exactly what I was thinking, but it's an option. Let's call it Option A. I'd like to hear feedback, and will be thinking about other options as well, including not hitting them in the face with it right away. This is probably the most up front one of the options I was thinking about.

![reactotron-usage](https://user-images.githubusercontent.com/1479215/42546542-abc07d80-8472-11e8-9eb7-4351402ca1d2.jpg)

rmevans9:
@skellock have you considered if you want to build a custom little analytics server to ship this data to or would you want to use Google Analytics or something of that nature?

skellock:
whatever yall think is best

jamonholmgren:
I'd probably vote for a small serverless app to pump this info into.

rmevans9:
@jamonholmgren sounds great. Given that this serverless app wouldn't technically be a part of Reactotrons offering would you think we build it as a standalone "analytics server" that we could ship analytics of multiple apps to (for ex Ignite)?

jamonholmgren:
That's an interesting idea, @rmevans9! I like it a lot.

markrickert:
This will be closed by #1406 when it lands.

PR #1406, opened 2024-01-18
Add basic analytics to Reactotron app with `react-ga4`

markrickert:
This adds google analytics to the reactotron app for basic feature and usage metrics. It's implemented using [`react-ga4`](https://github.com/codler/react-ga4) but the codebase can easily be switched to another service is we want.

The first time the user opens the app after upgrading to this version, they'll be presented with this screen:

Screenshot 2024-01-17 at 4 02 39 PM

Clicking the red "NO" button changes their user setting to opt-out of anonymous analytics. A user that opens the app and clicks "NO" should have zero analytics calls send to google-analytics.

Clicking the green "YES" button saves the user setting and starts allowing tracking calls to be sent to google analytics.

We are tracking the following things:

1. Screen views (with a hook that just listens to the path of `react-router`)
2. Interface button click events
3. Keyboard shortcut usage
4. Custom command usage statistics including the title of the custom command that was run.

Users can opt in or out of this feature by going to Reactotron -> Settings (which opens the settings json file in their default editor.

Screenshot 2024-01-17 at 4 16 41 PM

Editing this file to have the `analyticsOptOut` be `"true"` or `"false"` and then restarting the app will turn this setting on and off.

Screenshot 2024-01-17 at 4 18 42 PM

Things that we need to do before this can move out of draft status:

- [ ] Create a new google analytics property for Reactotron in our master InfiniteRed account and change the `GA4_KEY` variable.
- [ ] Look for other things that could benefit from usage tracking
- [ ] Brainstorm a way to track plugin usage (redux, mst, mmkv, etc)
- [ ] Validate verbiage for the Opt-in/out screen.


apps/reactotron-app/src/renderer/util/analyticsHelpers.tsx
```diff
@@ -0,0 +1,261 @@
...
+  useEffect(() => {
+    sendPageViewAnalyticsEvent(location.pathname)
+  }, [location, sendPageViewAnalyticsEvent])
+}
+
+// This is a custom alert that we use to ask the user if they want to opt-in to analytics
+// We use this instead of the default alert because we want to style it to match our app.
+// We inherit the styles from react-confirm-alert and override them as needed.
+// Unfortunately, we can't use styled-components here because react-confirm-alert doesn't support it.
+// This also means we have to hard-code the colors here instead of using our theme, which is not ideal.
```
Jpoliachik, line 178: just curious - how much is `react-confirm-alert` doing for us here? 
wondering if its worth the dependency if its difficult to style. could be nice to have full control of the UI, but might not be worth the effort.

Jpoliachik (commented):
Code looks good, nice work! 👍 

I think its worth a UX discussion on the impact of showing that alert on startup.
I suppose some other dev tools do something similar - like this setting in Android Studio: (I don't remember if you get prompted this on first startup or not)
image

IMO the alert could also use a UI pass if Justin has some bandwidth

apps/reactotron-app/src/renderer/pages/reactNative/Overlay.tsx
```diff
@@ -123,8 +125,18 @@ function Overlay() {
...
+        action: "OverlayDropImage",
+        label: "Too many files",
+      })
       return
     }
+    sendAnalyticsEvent({
+      category: "overlay",
+      action: "OverlayDropImage",
+      label: "Success",
+    })
```
jonmajorc, line 139: I noticed that each `sendAnalyticsEvent` has string values for category, action, and label. It may be worth using an enum for these constants. Then it's type-safe and easier to see what is available per key! Thoughts?

jonmajorc (commented):
This looks good! just left a suggestion 💯

apps/reactotron-app/src/renderer/images/index.ts
```diff
@@ -1,3 +1,4 @@
 export const reactotronLogo: string = require("./Reactotron-128.png").default
+export const reactotronAnalytics: string = require("./Reactotron-analytics.png").default
```
morganick, line 2: Let's throw this over to Justin and see what he'd like graphic wise.

apps/reactotron-app/src/renderer/App.tsx
```diff
@@ -28,6 +29,22 @@ const AppContainer = styled.div`
...
+// This wrapper container is used to track page views within the app automatically using react-router-dom
+// as well as to initialize analytics on app load.
+const AppContainer: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+  usePageTracking()
+  const { initializeAnalytics } = useAnalytics()
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      initializeAnalytics()
+    }, 250)
```
morganick, line 41: What are we waiting for here?

morganick (commented):
Just a couple of comments. Appears to work as expected.

Low GAFO: I find myself parsing this statement incorrectly half the time:
```
[analytics] user has changed opt-out status false
```

The status could be `unknown` | `opted-in` | `opted-out`. Then the statement could be "user has changed status to `opted-in`"

apps/reactotron-app/src/renderer/App.tsx
markrickert: While developing, I found that there was the possibility of this component rendering a couple times. This delay with the clearTimeout is meant to only initialize the analytics once the electron renderer process has completely opened the app and react has stopped reacting. Not sure if there's a better way to handle this other than some memoizing

markrickert:
I also found this discussion from a VERY long time ago #723 where @jamonholmgren had already mocked up a screen and verbiage.

apps/reactotron-app/src/renderer/util/analyticsHelpers.tsx
markrickert: Yeah, when i put it in the app, i was thinking it would be useful as-is, but the more i played with it, the more customizations I wound up doing. I'll probably rip this out before i change this pr to no longer be a draft.

apps/reactotron-app/src/renderer/util/analyticsHelpers.tsx
```diff
@@ -0,0 +1,261 @@
...
+      storeWatcher().removeAllListeners()
+    }
+  }, [])
+
+  // Get the user's opt-out status from the config store
+  const initializeAnalytics = () => {
+    const status = configStore.get("analyticsOptOut") as IOptOutStatus
+
+    if (status === "unknown") {
+      console.log(`[analytics] user has not opted in or out`)
```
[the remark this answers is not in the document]
markrickert: I think it's safe to leave in. It'll only ever console log once at the beginning of the app boot. It's useful for development to leave it in.

apps/reactotron-app/src/renderer/util/analyticsHelpers.tsx
```diff
@@ -0,0 +1,261 @@
...
+    }
+  }
+
+  // Initialize analytics and set some system data like the app version and platform
+  // as well as the mode we are running in. We don't want to send analytics events
+  // during tests, so we disable them if we are running in test mode.
+  // We also disable analytics if the user has opted out.
+  useEffect(() => {
+    const initialize = () => {
+      const testMode = process.env.NODE_ENV === "test" // we don't want to send analytics events during tests
```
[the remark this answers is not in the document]
markrickert: Well, i *did* want it to log during development so i could see that it was working 😄 

Good call to turn it off in DEV mode before we merge.

Screenshot 2024-01-24 at 9 57 53 AM

apps/reactotron-app/src/renderer/util/analyticsHelpers.tsx
```diff
@@ -0,0 +1,261 @@
...
+import packageJson from "../../../package.json"
+import { useLocation } from "react-router"
+import configStore from "../config"
+import { confirmAlert } from "react-confirm-alert" // Import
+import { reactotronAnalytics } from "../images"
+import "react-confirm-alert/src/react-confirm-alert.css" // Import css
+
+// This is the Google Analytics 4 key for Reactotron
+// TODO: Change this to the correct key for production.
+const GA4_KEY = "G-WZE3E5XCQ7"
```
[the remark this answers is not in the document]
markrickert: I don't think google analytics keys are considered private or sensitive information

apps/reactotron-app/src/renderer/pages/reactNative/Overlay.tsx
markrickert: Thanks for the suggestion @jonmajorc! Yeah, i need to clean up the types for sending events so we know exactly what we're going to send to the analytics server.

apps/reactotron-app/src/renderer/util/analyticsHelpers.tsx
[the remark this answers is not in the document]
morganick: It was less about privacy and more about being able to use CI to build it for the production key and have a dev key locally.

morganick:
@markrickert I rebased this on top of `master`. One thing I noticed when I was working with it was that setting the `analyticsOptOut: "unknown"` causes the application to break. It appears that's the default value and what should trigger the dialog to appear. Instead I get errors in console and a blank screen. I tested this both before and after rebasing master into this branch.

PR #1410, opened 2024-01-18
chore(tests): Migrate from @testing-library/react-hooks to @testing-library/react

markrickert:
Everything should be on react 18 now, so i swapped out `@testing-library/react-hooks` with `@testing-library/react` and updated all the imports.

Documentation for this change: https://github.com/testing-library/react-hooks-testing-library/blob/chore/migration-guide/MIGRATION_GUIDE.md

Tests passed locally.


morganick (approved):
LGTM 👍 and works as expected. All tests ✅

PR #1414, opened 2024-01-23
chore(docs): Add compose import from redux

markrickert:
Our docs have signifigantly changed since #1198 was opened so this will close #1198 with the same fix.

Thanks for the contribution, @tylerablake and sorry it took us so long to fix that!


frankcalise (approved):
🥇

PR #1415, opened 2024-01-23
chore(docs): Fix a yarn command problem.

markrickert:
This reimplements #1230 because the docs structure has significantly changed since that PR was opened.


frankcalise (approved):
👍

PR #1416, opened 2024-01-23
chore(ci): Update node orb and semantic-release versions

markrickert:
Moves `semantic-release` dependency from `reactotron-react-js` to the root package.json file and updates it to the latest version.

Closes #1261

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "eslint-plugin-standard": "^5.0.0",
     "nx": "17.0.3",
     "prettier": "^3.0.3",
+    "semantic-release": "^23.0.0",
     "typescript": "^4.9.5",
     "zx": "^7.2.3"
   },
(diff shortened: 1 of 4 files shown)


frankcalise (approved):
Looks good!

package.json
```diff
@@ -43,6 +43,7 @@
     "eslint-plugin-standard": "^5.0.0",
     "nx": "17.0.3",
     "prettier": "^3.0.3",
+    "semantic-release": "^23.0.0",
```
joshuayoes, line 46: We can remove `semantic-release` all together actually!

joshuayoes (approved):
Nice update here, only a little comment that we can simplify this even further.

package.json
markrickert: oh great! I'll do that then :D

PR #1424, opened 2024-01-26
feat(example-app): Add a LOT of functionality to the demo app

markrickert:
This work builds on the old demo app that was removed a while ago.

@morganick added the basic app framework the other day and this PR fleshes out the app to be more Reactotron branded. There's still some work to do particularly fleshing out the custom commands screen and getting some interaction there. Maybe we navigate them to a screen that has no way to go back, but instruct them to run a command in Reactotron to fix the navigation state or something.

I think that this will continue to be a work in progress. I really like having a "kitchen sink" app that demo's all of Reactotron's functionality.

Check out this video for how it looks in action:

https://github.com/infinitered/reactotron/assets/139261/d78f1782-579b-45c4-bbb7-1959d042e5e9

# Here's some screenshots: 

<table>
  <tr>
    <td></td>
    <td></td>
<td>
</td>
   </tr> 
</table>


apps/example-app/app/screens/LoggingScreen.tsx
```diff
@@ -0,0 +1,171 @@
...
+            image: { uri },
+          })
+        }
+      },
+      (error) => console.error("Oops, snapshot failed", error),
+    )
+  }
+
+  return (
+    <Screen style={$container} preset="scroll">
```
morganick, line 33: Can we add safe area padding to the bottom of this screen.

apps/example-app/app/screens/LoggingScreen.tsx
```diff
@@ -0,0 +1,171 @@
...
+                })
+              }
+            }}
+          />
+          <Button
+            text="Log a Screenshot"
+            textStyle={$darkText}
+            style={$button}
+            onPress={handleScreenshot}
+          />
```
morganick, line 106: Appears that this does not respect dark mode as my screenshot has a white background instead of a black one.

apps/example-app/package.json
```diff
@@ -84,7 +91,7 @@
     "eslint-plugin-promise": "6.0.0",
     "eslint-plugin-react": "7.30.0",
     "eslint-plugin-react-native": "4.0.0",
-    "eslint-plugin-reactotron": "^0.1.1",
+    "eslint-plugin-reactotron": "workspace:*",
```
morganick, line 94: 🙌

morganick (commented):
Just a couple of comments and then g2g

apps/example-app/app/devtools/ReactotronClient.web.ts
```diff
@@ -3,6 +3,9 @@
  * of Reactotron's client.
  *
  * React Native is loaded from ReactotronClient.ts.
+ *
+ * If your project does not need web support, you can delete this file and
+ * remove reactotron-react-js from your package.json dependencies.
```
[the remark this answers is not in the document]
markrickert: You're gonna have to take this one up with ignite 😉  https://github.com/infinitered/ignite/pull/2617

apps/example-app/app/navigators/AppNavigator.tsx
```diff
@@ -47,20 +55,68 @@ export type AppStackScreenProps<T extends keyof AppStackParamList> = NativeStack
...
-    <Stack.Navigator screenOptions={{ headerShown: false, navigationBarColor: colors.background }}>
+    <Stack.Navigator
+      screenOptions={{ headerShown: false, navigationBarColor: colors.background }}
+      initialRouteName="Welcome"
+    >
       <Stack.Screen name="Welcome" component={Screens.WelcomeScreen} />
+      <Stack.Group
+        screenOptions={{
+          headerShown: true,
+          // headerTransparent: true,
```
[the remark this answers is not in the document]
markrickert: Yeah, that was some leftover code from trying to make the linear gradient in the title bar.

apps/example-app/app/screens/LoggingScreen.tsx
markrickert: I think it's because the background color is on the top container and not the View within the `<ViewShot>`. I'll fix.

morganick:
Looks like there is a formatting issue. Likely due to the prettier config change.

PR #1425, opened 2024-01-27
Added Electron BrowserWindow title

therealrinku:
Tested on Debian 12 OS

before:
before_name_change

after :
after_name_change


markrickert:
Nice! Thanks for the contribution. I'm going to override the CI on this one because i don't see any issues.

Issue #1282, opened 2023-03-19
Add HTTP Status Code to API Response Records


matthewmturner:
If you have a general question, please ask it in the #reactotron channel of the
[Infinite Red Community slack](https://community.infinite.red/).

When reporting a bug, please include steps to reproduce the issue.

It would be nice and much easier to sort through the timeline if we could see the HTTP response code on API Response records without having to click into the record. i.e. have `API RESPONSE (200)` instead of just `API RESPONSE`

PR #1434, opened 2024-02-05
Show response code in response records

therealrinku:
Addressing Issue #1282 

Screenshot from 2024-02-05 23-47-03


markrickert:
This is a great addition to the network timeline commands! Thanks for doing this and opening and PR! I cloned your branch and rebased on master and everything looks good!

PR #1439, opened 2024-02-09
chore(tools): Update a lot of development workflow packages and move to modern  & consistent `rollup` configuration

markrickert:
This PR updates a lot of behind-the-scenes tools like `nx` and `rollup`.

I completely replaced the old 1.x and 2.x rollup configuration and old deprecated plugins with their latest and greatest versions.

rollup config files were all renamed to `rollup.config.cjs` and now use the `@rollup/plugin-typescript` instead of the old babel integration.

I also updated `@jscutlery/semver` and related dependencies to the latest version as well. There's a small risk that this might break CI, but 🤞


markrickert:
Of note, there are some `electron` minor package updates in this PR as well.

joshuayoes (approved):
I did a spot of the `/dist` output and installed all the packages into an Ignite app as well, it all looks good!

markrickert:
Looks like nothing broke!

markrickert:
I was wrong. 😭

PR #1440, opened 2024-02-13
fix: configureStore instructions for redux

siddarthkay:
fix docs to get `reactotron` from `ReactotronConfig` instead of `reactotron-react-native`


markrickert (approved):
Thanks for this! I'll merge and publish when #1442 is merged.

PR #1442, opened 2024-02-13
chore(monorepo): Remove all prettier config files and inherit from the root.

markrickert:
All the prettier config files were basically the same in every directory so I reworked the monrepo to only have a top-level `.prettierrc` and `.prettierignore` files.

This also happens to fix the problem where `symver` writes to the changelog and prettier doesn't like the format that `symver` produces.


joshuayoes (approved):
Nice way to reduce files 👌

PR #1457, opened 2024-03-21
fix(reactotron-core-ui): Overlapping timeline layout

opakholis:

## Describe your PR

When we have a long request parameter, it appears that the layout overlaps.

Please take a look at the following demo. 

Before:
CleanShot 2024-03-21 at 6 19 25

After:
CleanShot 2024-03-21 at 6 39 19

### DIFF
--- a/lib/reactotron-core-ui/src/components/TimelineCommand/index.tsx
+++ b/lib/reactotron-core-ui/src/components/TimelineCommand/index.tsx
@@ -80,6 +80,7 @@ const ChildrenContainer = styled.div`
   animation: fade-up 0.25s;
   will-change: transform opacity;
   padding: 0 40px 30px 40px;
+  word-break: break-word;
 `
 
 interface Props {


lib/reactotron-core-ui/src/components/TimelineCommand/index.tsx
```diff
@@ -80,6 +80,7 @@ const ChildrenContainer = styled.div`
   animation: fade-up 0.25s;
   will-change: transform opacity;
   padding: 0 40px 30px 40px;
+  word-break: break-word;
```
morganick, line 83: @opakholis thanks for jumping in to help fix this. `break-word` is deprecated so let's use the same `word-break` strategy as we use in the `PreviewContainer` on line 61 here:
```suggestion
  word-break: break-all;
```

morganick (changes requested):
@markrickert the issue has to do with where we display the URL of the API request. By applying the `word-break: break-all` to the parent container it is inherited by all the children.

To reproduce, you can make an API call with the example application and edit the DOM to replace the URL with the following:
```
https://jsonplaceholder.typicode.com/todos/1?random=1710343976e11707d502a743efb45e2cf4d2036f7deKdgIyq1GR4Flb%2FxzNNEcRhnTcfoWgoSF%2BEZH%2Fjm%2FEE%3D
```

This fixes this particular instance of the issue, but we should probably enforce that the container holds its boundaries properly higher up.

lib/reactotron-core-ui/src/components/TimelineCommand/index.tsx
opakholis: @morganick thanks for the detailed explanation and suggestion. I didn't realize that `break-word` was deprecated.

morganick: @opakholis I didn't either until I looked it up to make sure that's the one we wanted. 😅

morganick (approved):
LGTM 👍 and works as expected.

morganick:
Thanks @opakholis 🙌

PR #1459, opened 2024-03-22
Fix AsyncStorage import for react-native

nkleidis:
According to the [documentation](https://react-native-async-storage.github.io/async-storage/docs/usage) of AsyncStorage the import should be the default import


markrickert:
Indeed! You are correct! Thanks for catching this!

PR #1464, opened 2024-03-28
TRON-10: Async Storage not Required

morganick:

## Describe your PR
When an application does not have Async Storage or other persistent storage mechanism configured, Reactotron will create a new connection and device each time the application is reloaded.

### Reproduce Behavior
Use the example application inside of the project and comment out line 8:
```
// import AsyncStorage from "@react-native-async-storage/async-storage"
```

and line 36:
```
// reactotron.setAsyncStorageHandler?.(AsyncStorage)
```

This will deactivate Async Storage and no longer persist the generated unique `clientId` coming from the websocket server.

### Research & Constraints
* Reactotron currently works with Expo Go. In order to maintain that support, we cannot use packages like `react-native-device-info` as it has native components that would need to be installed.
* `expo-device` did not provide any additional functionality on top of what we already had access to.
* Websocks do have a unique key, but we'd have to store that key on the device...
* Device orientation should not change `clientId`. (thanks @frankcalise)
* Switching applications on the same simulator should not mingle their timelines.

### Solution
When Async Storage or the like is not configured, generate a device fingerprint as a _best effort_ to keep the connected devices stable. This seems like a good enough solution as there are no additional dependencies and working with information that we already had.
| Before | After |
|--------|--------|
| <video src="https://github.com/infinitered/reactotron/assets/151139/5c18642b-121f-41cf-9864-e32a1019dc1a" /> | <video src="https://github.com/infinitered/reactotron/assets/151139/264eedfd-65e8-4781-9d5d-d55006138b53" /> |

## Additional Notes
Do we have documentation in Reactotron for configuring MMKV or Expo SecureStore?

### DIFF
--- a/lib/reactotron-react-native/src/reactotron-react-native.ts
+++ b/lib/reactotron-react-native/src/reactotron-react-native.ts
@@ -71,15 +71,19 @@ const DEFAULTS: ClientOptions<ReactotronReactNative> = {
     }
 
     // Generate clientId based on the device info
-    const dimensions = getReactNativeDimensions()
+    const { screenWidth, screenHeight, screenScale } = getReactNativeDimensions()
+
+    // Accounting for screen rotation
+    const dimensions = [screenWidth, screenHeight].sort().join("-")
+
     tempClientId = [
       name,
       Platform.OS,
       Platform.Version,
+      constants.systemName,
       constants.Model,
-      dimensions.screenWidth,
-      dimensions.screenWidth,
-      dimensions.screenScale,
+      dimensions,
+      screenScale,
     ]
       .filter(Boolean)
       .join("-")
--- a/lib/reactotron-react-native/src/reactotron-react-native.ts
+++ b/lib/reactotron-react-native/src/reactotron-react-native.ts
@@ -65,10 +65,24 @@ const DEFAULTS: ClientOptions<ReactotronReactNative> = {
     ...getReactNativeDimensions(),
   },
   /* eslint-disable @typescript-eslint/no-use-before-define */
-  getClientId: async () => {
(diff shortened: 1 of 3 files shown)


markrickert (approved):
This is awesome! This should fix my issue in a non-async-storage project?

Screenshot 2024-03-28 at 12 54 53 PM

lindboe:
Trying this out today!

lib/reactotron-react-native/src/reactotron-react-native.ts
```diff
@@ -65,11 +65,29 @@ const DEFAULTS: ClientOptions<ReactotronReactNative> = {
...
+  getClientId: async (name: string = "") => {
     if (reactotron.asyncStorageHandler) {
       return reactotron.asyncStorageHandler.getItem(REACTOTRON_ASYNC_CLIENT_ID)
     }
 
+    // Generate clientId based on the device info
+    const { screenWidth, screenHeight, screenScale } = getReactNativeDimensions()
+
+    // Accounting for screen rotation
+    const dimensions = [screenWidth, screenHeight].sort().join("-")
```
frankcalise, line 77: u fancy

morganick: Yeah... overkill really. I originally had the scale in there too, but didn't like the way that it read. 😅

Might be better to do '${smaller}x${larger}' with a ternary.

morganick:
@lindboe have you had a chance to check this out?

lindboe:
@morganick Sorry, not yet, priorities changed. I'll try to find some time, but don't block on me, can always follow up later

Issue #1398, opened 2024-01-09
iOS Release Build app crash on real device


Ajmal0197:
`Terminating app due to uncaught exception 'RCTFatalException: Unhandled JS Exception: Error: invalid host, js engine: hermes', reason: 'Unhandled JS Exception: Error: invalid host, js engine: hermes`

App crashing in ios release mode.
Refer for detail:
https://github.com/facebook/hermes/issues/1228

GSFZamai:
Same here. In my case only on iOS Release Mode

rodrigorethink:
The problem started to happen in version 5.0.4 of reactotron-react-native. Using version 5.0.3 the crash does not occur.
Comparing the versions, it can be seen that the getHost routine in the file /reactotron-react-native/src/reactotron-react-native.ts uses a simplified version, which works for Expo applications, but in iOS builds in release mode this new version uses the URI of the embedded bundle.

Example:
using the code NativeModules?.SourceCode?.getConstants().scriptURL in release mode you get something like file:///private/var/containers/Bundle/Application/38A07257-A8E8-4565-BBE6-6B874A82A47B/app_name.app /main.jsbundle
The new getHost routine will extract "private" and will try to use it as a host, which generates the "invalid host" exception

GSFZamai:
> Hi @rodrigorethink Iam using reacctotron version 5.0.3 only but still its getting crashed!!

I've downgraded the version and worked for me.
Did you try to remove node_modules and install all again?

lachu97:
Hi @GSFZamai Yes, i have tried it but still in the production the same thing happens...

rodrigorethink:
> Hi @GSFZamai Yes, i have tried it but still in the production the same thing happens...

I had to remove node_modules, clear metro and XCode cache to work locally after downgrade

GSFZamai:
> Hi @GSFZamai Yes, i have tried it but still in the production the same thing happens...

You can try the steps done by @rodrigorethink.

Can also try delete ~/Library/Developer/Xcode/DerivedData.
It Might be some cached build.

Verify your yarn.lock to ensure thet the version installed is really 5.0.3
![image](https://github.com/infinitered/reactotron/assets/47357785/1954aa27-441c-4ba4-a5ed-4cf0e9cf4bce)

luanlcampos:
In my case, I also had to downgrade `reactotron-redux` to version 3.1.3.

lachu97:
[...]

PR #1468, opened 2024-04-04
Update code snippet using require instead of import.

alexsegura:
Since #1398 we know that `require()` should be used, I updated the code snippet so that people can copy/paste working code.


markrickert:
Thanks for updating this!

PR #1426, opened 2024-01-29
New Network Page(demo)

meftunca:
Because I need it, I created an additional page where I can see network requests more easily. We can integrate this into the project by making it more quality.
Screenshot 2024-01-29 at 16 10 18(1)

PR #1471, opened 2024-04-20
feat(reactotron-app): Network tab with api responses

finalquest:

- [ x] `yarn build-and-test:local` passes
- [ x] I have added tests for any new features, if relevant

## Describe your PR
I made a new network tab to group the api.responses messages.
Made a custom command to clear all the commands that are/not are api. 
Removed the api filter from the Timeline modal

## Next Steps
This changes only moved the rendered cell to another list.
I think that the command list should be splitted in commands and networkCommands.
The current solution works over the full command list.
Maybe we should make a right sidebar with all the header/request/response. Something like Flipper have

### DIFF
--- a/apps/reactotron-app/src/renderer/App.tsx
+++ b/apps/reactotron-app/src/renderer/App.tsx
@@ -15,6 +15,7 @@ import Overlay from "./pages/reactNative/Overlay"
 import Storybook from "./pages/reactNative/Storybook"
 import CustomCommands from "./pages/customCommands"
 import Help from "./pages/help"
+import NetworkPage from "./pages/network"
 
 const AppContainer = styled.div`
   position: absolute;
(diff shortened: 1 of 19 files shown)


GrinZero:
Hello 👋 Do you have a preview image of the effect

jzaefferer:
Looks like this is a duplicate of https://github.com/infinitered/reactotron/pull/1426 though.

jamonholmgren:
I do like the idea here, @finalquest, but since it looks like you're pulling in #1426 and building on that, we should leave their commits in to credit them for the work they've done.

Regardless, I'm not sure that I'll accept this as-is. Will do some thinking about this. I think it's a good idea, but the implementation here might need some more thought.

longb1997:
This is really needed, please take a look

PR #1476, opened 2024-05-02
Issue-1473: App Crash with Invalid scriptURL

### LINKED ISSUE #1473: Reactotron will crash first app runs on real iOS devices
### Describe the bug

On the first run of an RN app on a real iOS device (dev build), Reactotron will throw `invalid host`, which can lead to crashing the app (here because we were trying to add the reactotron redux enhancer to the store, but that failed, so the store failed to be created, crashing the app).

I think the underlying cause of this particular crash is because we need to click "Allow" to find devices on the local network. After pressing "Allow" and reloading, the app works fine.

In addition, Reactotron should be more resilient to errors like this and avoid crashing apps. It should:
1. Ensure that any auto-generated hostname or other config never fails the validate test for hostname
2. Ensure `validate` tests don't crash the app if they fail. Utilities that depend on Reactotron plugins would ideally return a no-op version (e.g., the redux enhancer) so that the app can continue to work.
3. Errors from `validate` should make it more clear that they're from Reactotron in the message

Notes:
1. This may depend on which version of RN you're on. I saw it on RN 71, but I do see the underlying code to get the source code URL in RN has been changed since. This could mean differences in whether "permission" is needed?
2. I don't know of a simple way to "wait" until "Allow" has been pressed, so the fix would be to ensure we don't crash the app and then also notify users they'll need to reload the app for it to connect to Reactotron (with docs or a yellow box warning)

reactotron-react-native: 5.1.6

### Reactotron version

3.7.0

morganick:

## Describe your PR

Fixes #1473 where calling split on an invalid string throws a `TypeError`.

In the event that the following line does not provide us with a valid URL:
https://github.com/facebook/react-native/blob/b38f80aeb6ad986c64fd03f53b2e01a7990e1533/packages/react-native/React/CoreModules/RCTSourceCode.mm#L38

We should catch this and warn the user that we are falling back to the default host:
```
WARN  getHost: "Invalid URL: null" for scriptURL - Falling back to localhost
```

## How to Reproduce

I was unable to reproduce this issue locally with my devices as it would not crash the application, but just throw the error. Due to the circumstances of the original issue and that applications configuration, this should allow that application to continue to boot.

## Additional Notes

It is still possible to configure Reactotron with an invalid host. This will equally throw an error of `invalid url`. I'd like to create a new PR that allows the application to continue to function, keeps Reactotron is a disconnected state, and notifies the user.

### DIFF
--- a/lib/reactotron-react-native/src/helpers/parseURL.test.ts
+++ b/lib/reactotron-react-native/src/helpers/parseURL.test.ts
@@ -11,6 +11,7 @@ describe("getHostFromUrl", () => {
     Object.entries({
       localhost: "localhost",
       "127.0.0.1": "127.0.0.1",
+      "[::1]": "[::1]",
     }).forEach(([host, url]) => {
       expect(getHostFromUrl(url)).toEqual(host)
     })
@@ -51,6 +52,17 @@ describe("getHostFromUrl", () => {
         "http://localhost:8081/.expo/.virtual-metro-entry.bundle?platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.reactotronapp",
       "192.168.1.141":
         "https://192.168.1.141:8081/.expo/.virtual-metro-entry.bundle?platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.reactotronapp",
+    }).forEach(([host, url]) => {
+      expect(getHostFromUrl(url)).toEqual(host)
+    })
+  })
+
+  it("should get the host from an IPv6 URL and ignore path, port, and query params", () => {
+    Object.entries({
+      "[::1]":
+        "http://[::1]:8081/.expo/.virtual-metro-entry.bundle?platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.reactotronapp",
+      "[2001:0db8:85a3:0000:0000:8a2e:0370:7334]":
+        "https://[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:8081/.expo/.virtual-metro-entry.bundle?platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.reactotronapp",
     }).forEach(([host, url]) => {
       expect(getHostFromUrl(url)).toEqual(host)
     })
--- a/lib/reactotron-react-native/src/helpers/parseURL.test.ts
+++ b/lib/reactotron-react-native/src/helpers/parseURL.test.ts
@@ -37,15 +37,6 @@ describe("getHostFromUrl", () => {
     })
   })
 
-  it("should get the host from URL with expo scheme for Expo Go", () => {
-    Object.entries({
-      localhost: "exp://localhost",
-      "example.com": "exp://example.com",
-    }).forEach(([host, url]) => {
-      expect(getHostFromUrl(url)).toEqual(host)
-    })
-  })
-
   it("should get the host from URL and ignore path, port, and query params", () => {
     Object.entries({
       localhost:
--- a/lib/reactotron-react-native/src/helpers/parseURL.test.ts
+++ b/lib/reactotron-react-native/src/helpers/parseURL.test.ts
@@ -36,6 +36,15 @@ describe("getHostFromUrl", () => {
     })
   })
 
+  it("should get the host from URL with expo scheme for Expo Go", () => {
+    Object.entries({
+      localhost: "exp://localhost",
+      "example.com": "exp://example.com",
+    }).forEach(([host, url]) => {
+      expect(getHostFromUrl(url)).toEqual(host)
+    })
+  })
+
   it("should get the host from URL and ignore path, port, and query params", () => {
     Object.entries({
       localhost:
--- a/lib/reactotron-react-native/src/helpers/parseURL.test.ts
+++ b/lib/reactotron-react-native/src/helpers/parseURL.test.ts
@@ -1,0 +1,59 @@
+import { getHostFromUrl } from "./parseURL"
+
+describe("getHostFromUrl", () => {
+  it("should throw when no host is found", () => {
+    expect(() => {
+      getHostFromUrl("")
+    }).toThrow()
+  })
+
+  it("should get host from URL without scheme", () => {
+    Object.entries({
+      localhost: "localhost",
+      "127.0.0.1": "127.0.0.1",
+    }).forEach(([host, url]) => {
+      expect(getHostFromUrl(url)).toEqual(host)
+    })
+    expect(getHostFromUrl("localhost")).toEqual("localhost")
+    expect(getHostFromUrl("127.0.0.1")).toEqual("127.0.0.1")
+  })
+
+  it("should get the host from URL with http scheme", () => {
+    Object.entries({
+      localhost: "http://localhost",
+      "example.com": "http://example.com",
+    }).forEach(([host, url]) => {
+      expect(getHostFromUrl(url)).toEqual(host)
+    })
+  })
+
+  it("should get the host from URL with https scheme", () => {
+    Object.entries({
+      localhost: "https://localhost",
+      "example.com": "https://example.com",
+    }).forEach(([host, url]) => {
+      expect(getHostFromUrl(url)).toEqual(host)
+    })
+  })
+
+  it("should get the host from URL and ignore path, port, and query params", () => {
+    Object.entries({
+      localhost:
+        "http://localhost:8081/.expo/.virtual-metro-entry.bundle?platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.reactotronapp",
+      "192.168.1.141":
+        "https://192.168.1.141:8081/.expo/.virtual-metro-entry.bundle?platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.reactotronapp",
+    }).forEach(([host, url]) => {
+      expect(getHostFromUrl(url)).toEqual(host)
+    })
+  })
+
+  it("should get the host from URL with hyphens", () => {
+    expect(getHostFromUrl("https://example-app.com")).toEqual("example-app.com")
+  })
+
+  it("should throw when the URL is an unsupported scheme", () => {
+    expect(() => {
+      getHostFromUrl("file:///Users/tron")
+    }).toThrow()
+  })
+})
--- a/lib/reactotron-react-native/src/reactotron-react-native.ts
+++ b/lib/reactotron-react-native/src/reactotron-react-native.ts
@@ -36,7 +36,15 @@ let tempClientId: string | null = null
 const getHost = (defaultHost = "localhost") => {
   try {
     // RN Reference: https://github.com/facebook/react-native/blob/main/packages/react-native/src/private/specs/modules/NativeSourceCode.js
-    return new URL(NativeModules?.SourceCode?.getConstants().scriptURL).hostname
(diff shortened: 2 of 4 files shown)


lib/reactotron-react-native/src/reactotron-react-native.ts
```diff
@@ -33,13 +33,15 @@ let tempClientId: string | null = null
...
-  typeof NativeModules?.SourceCode?.getConstants().scriptURL === "string" // type guard in case this ever breaks https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/NativeModules/specs/NativeSourceCode.js#L15-L21
-    ? NativeModules.SourceCode.scriptURL // Example: 'http://192.168.0.100:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.helloworld'
-        .split("://")[1] // Remove the scheme: '192.168.0.100:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.helloworld'
-        .split("/")[0] // Remove the path: '192.168.0.100:8081'
-        .split(":")[0] // Remove the port: '192.168.0.100'
-    : defaultHost
+const getHost = (defaultHost = "localhost") => {
+  try {
+    // RN Reference: https://github.com/facebook/react-native/blob/main/packages/react-native/src/private/specs/modules/NativeSourceCode.js
+    return new URL(NativeModules?.SourceCode?.getConstants().scriptURL).hostname
```
morganick, line 39: Didn't see any reason not to use URL class here for this purpose:
https://github.com/facebook/react-native/blob/b38f80aeb6ad986c64fd03f53b2e01a7990e1533/packages/react-native/types/modules/globals.d.ts#L396

lib/reactotron-react-native/src/reactotron-react-native.ts
```diff
@@ -33,13 +33,15 @@ let tempClientId: string | null = null
...
-const getHost = (defaultHost = "localhost") =>
-  typeof NativeModules?.SourceCode?.getConstants().scriptURL === "string" // type guard in case this ever breaks https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/NativeModules/specs/NativeSourceCode.js#L15-L21
-    ? NativeModules.SourceCode.scriptURL // Example: 'http://192.168.0.100:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.helloworld'
-        .split("://")[1] // Remove the scheme: '192.168.0.100:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.helloworld'
-        .split("/")[0] // Remove the path: '192.168.0.100:8081'
-        .split(":")[0] // Remove the port: '192.168.0.100'
-    : defaultHost
+const getHost = (defaultHost = "localhost") => {
+  try {
+    // RN Reference: https://github.com/facebook/react-native/blob/main/packages/react-native/src/private/specs/modules/NativeSourceCode.js
```
morganick, line 38: This URL gives you more context as to what's available with this native module. I can change it back to the old link if that's more clear.

lib/reactotron-react-native/src/reactotron-react-native.ts
lindboe: IIRC Hermes URL implementation has some problems. Josh has commented on this: https://github.com/facebook/hermes/discussions/1072#discussioncomment-6578180. I'm not sure of the current state, but looking...

From https://www.npmjs.com/package/react-native-url-polyfill, they claim it doesn't work with 'localhost' properly (https://github.com/facebook/react-native/issues/26019), and creating new URLs can throw exceptions (https://github.com/facebook/react-native/issues/16434).

I think if we take this route we need to be sure to test it with a non-Expo app, since as that discussion mentions they added their own polyfill. Maybe test the oldest supported RN version too in case changes have been made that I haven't seen.

morganick: image
Looks like the Expo polyfill was doing its job. The above screenshot is from a fresh react native cli app. I'll switch it out with something that handles just this case.

morganick: I was unable to use the same package that Expo is using for their URL Here's the scoop:

We use rollup for building the core dependencies into the packages like `reactotron-react-native`. I was trying to use a package that uses commonjs exports inside of the library. Evidently, Rollup has deprecated the way that is handled and moved it into another package: https://rollupjs.org/troubleshooting/#error-name-is-not-exported-by-module ( I was specifically running into this issue)

I could install that package in any other application, but not for use in our monorepo because of rollup... but I didn't figure that out until this morning because after I removed it... `reactotron-react-native` would no longer build. Even after reverting all of the changes calling yarn clean and yarn reset. It wasn't until I switched branches back to `master`... where it started to build again. Then I switched back to my branch without the new package and it continued the build 😱

So I was then trying to figure out what cache is not being cleared. (I didn't realize rollup was involved.) Come to find out that I believe the nx cache is cleared when you switch branches and code changes in that dep. So in order for me to reliably add that package and not foobar the build I would have to run `yarn nx reset && yarn reset`.

I have updated our `yarn reset` script for this reason to include a call to `yarn nx reset`.

I moved this to being a regex for now until we can do the prerequisite Rollup upgrades.

lib/reactotron-react-native/src/reactotron-react-native.ts
```diff
@@ -33,13 +33,23 @@ let tempClientId: string | null = null
...
+    const scriptURL = NativeModules?.SourceCode?.getConstants().scriptURL
+    if (typeof scriptURL !== "string") throw new Error("Invalid non-string URL")
+
+    // Using a capture group to extract the hostname from a URL
+    const host = scriptURL.match(/^(?:https?:\/\/)?([^/:\s]+)(?::\d+)?(?:[/?#]|$)/)?.[1]
```
morganick, line 43: Test cases: https://regex101.com/r/WIHVuQ/2

scripts/reset.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+echo "Nx Reset - Clears all the cached Nx artifacts and metadata about the workspace and shuts down the Nx Daemon."
+yarn nx reset
+
```
morganick, line 5: @joshuayoes https://nx.dev/nx-api/nx/documents/reset

morganick:
@lindboe I have validated this with Expo apps on my devices, simulator, and emulator for both iOS and Android. I also validated this on device Android and iOS simulator with RN 0.74 app using RN cli.

Could you take this for a spin inside your application on your devices to make sure we've covered the cases we need to.

lib/reactotron-react-native/src/reactotron-react-native.ts
morganick: Future 🤔 ...

We could try to use the built in `URL` class if it has the `host` or `hostname` implemented. This would work in current Expo applications. We could then fall back to the regex if that's not available.

Reasons:

* Using what's there so it stays in line with what the application is already using
* More robust support for URLs - The regex above does not support IPv6.
* Hope that URL parsing gets fully implemented so that it's available across the board without polyfill or external packages and then we can remove the regex.

For now, we'll utilize the regex until we can either import the standardized URL parser or it becomes a first class citizen of the platform. To @lindboe point, having one solution means the potential bug reports will be simpler to triage.

joshuayoes: Could we put these regex in the codebase as unit tests? That way we can add more tests if we have more URL issues in the future.

joshuayoes (approved):
This looks like a good solution! I left one comment about adding unit tests I'd like to see but this PR could be merged in as is

lib/reactotron-react-native/src/reactotron-react-native.ts
lindboe: One thing I like to do with regexes is comments that explain their matching section-by-section (here it seems feasible to do per-capture group) just to make sure everybody's on the same page about what's intended

Unit tests aren't a bad idea either 😄

lindboe (approved):
Tested it out, looks like it works as intended.

migueldaipre:
Hey guys, any news on when this will be released?

lib/reactotron-react-native/src/helpers/parseURL.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  it("should throw when no host is found", () => {
+    expect(() => {
+      getHostFromUrl("")
+    }).toThrow()
+  })
+
+  it("should get host from URL without scheme", () => {
+    Object.entries({
+      localhost: "localhost",
+      "127.0.0.1": "127.0.0.1",
```
lindboe, line 13: do we need to match on ipv6?

lib/reactotron-react-native/src/helpers/parseURL.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  it("should get the host from URL with https scheme", () => {
+    Object.entries({
+      localhost: "https://localhost",
+      "example.com": "https://example.com",
+    }).forEach(([host, url]) => {
+      expect(getHostFromUrl(url)).toEqual(host)
+    })
+  })
+
+  it("should get the host from URL and ignore path, port, and query params", () => {
```
lindboe, line 39: Should we also test with Expo tunnel-style URLs? https://docs.expo.dev/more/expo-cli/#tunneling.

lib/reactotron-react-native/src/helpers/parseURL.test.ts
morganick: Lol, this was the 🐰 🕳️ I was trying to avoid 😜 I have an idea for tomorrow.

lindboe: Hey, fair enough, you can tell me you don't think we need it! I have no idea whether we do or not.

morganick: I've added IPv6 and `exp://` schema for Expo Go.

lib/reactotron-react-native/src/helpers/parseURL.test.ts
morganick: I'm not quite sure how to support tunneling honestly. I think you'd have to create a tunnel for each port that you wanted to connect to. This is something I'd probably reach for Tailscale for... which reminds me I should create a blog post for that. Either way, I think we'd still get the host right for it. Just would have to make sure we forwarded the right ports to Reactotron, yes?

lib/reactotron-react-native/src/helpers/parseURL.test.ts
morganick: `exp://` may not be needed, but I included it as when you run `expo start` it says that metro is waiting on `exp://192.168.1.x:8081` so if that comes through on the `scriptURL` we'll handle it.

I added specs for both IPv6 and `exp` schema.

lib/reactotron-react-native/src/helpers/parseURL.test.ts
lindboe: With this comment I only meant we should check we don't break anything that wasn't broken for Expo's tunneling support. I'm not sure if `exp://` is used there for hosting the JS bundle, I thought maybe it was just used for deep links now? Not sure.

morganick:
@joshuayoes I went full regex based on our conversation on Friday.

lib/reactotron-react-native/src/helpers/parseURL.test.ts
morganick: It appears to work with Expo Go without it. Removed.

morganick: @lindboe In order to support tunneling, I believe we'd have to have some way to know to include the port. This PR only replaces the old method of finding the host with safeguards around not crashing the application if the URL is not valid. Maybe we tackle that in another PR?

lindboe: Yeah that makes sense 👍

lindboe (approved):
Looks good!

PR #1477, opened 2024-05-06
fix(docs): Install AsyncStorage as dependency

carmeli-tamir:
Without AsyncStorage dependency, the instruction in "Step 2 - Initialization" simply doesn´t work if the developer app doesn´t already use AsyncStorage, since it does ´Reactotron.setAsyncStorageHandler(AsyncStorage)´

If the developer app already uses AsyncStorage, this is harmless.


morganick:
Hey @carmeli-tamir, it should work without Async Storage. However, I noticed late last week that I forgot to update the docs so that `setAsyncStorageHandler` was not part of the base configuration 😅 If you remove that from the config, it should work as you expect it without Async Storage. We didn't want you to have to have Async Storage installed in your app in order to use Reactotron.

Let me know if you run into further issue.

carmeli-tamir:
Thanks for the clarification. I´ve sumbitted a PR to remove AsyncStorage from the base config. https://github.com/infinitered/reactotron/pull/1479/files

PR #1488, opened 2024-07-03
chore(.tool-version): remove incorrect yarn version

joshuayoes:

## Describe your PR

The yarn version is handled via the `packageManager` key in the package.json
https://github.com/infinitered/reactotron/blob/fix/yarn-tool-version/package.json#L81

And it appears that [asdf-yarn](https://github.com/twuni/asdf-yarn) is only needed for yarn 1 versions, not later yarn. 

This PR removes the reference since [yarn 4](https://github.com/infinitered/reactotron/blob/fix/yarn-tool-version/.yarn/releases/yarn-4.1.1.cjs#L1) is already bundled into the repo.


morganick (approved):
LGTM 👍 Yep, I should have done this when I upgraded. 🤦

PR #1489, opened 2024-07-03
chore(package.json): remove --notify from "test:watch" script

joshuayoes:

## Describe your PR

This PR removes the `--notify` flag from all of the `test:watch` scripts due to a failing dependency on [node-notifier](https://www.npmjs.com/package/node-notifier). I personally would not want my tests results to be notified in my computer notifications, so I see removing this as a good behavior.

## Before
1. `yarn workspace reactotron-mst test:watch`
2. See error:
Screenshot 2024-07-03 at 2 57 45 PM

## After
1. `yarn workspace reactotron-mst test:watch`
2. See that test suite passes and goes into watch mode
Screenshot 2024-07-03 at 3 00 15 PM

### DIFF
--- a/apps/reactotron-app/package.json
+++ b/apps/reactotron-app/package.json
@@ -26,7 +26,7 @@
     "build:release": "node scripts/build.release.js",
     "release:artifacts": "node scripts/release.artifacts.js",
     "test": "jest",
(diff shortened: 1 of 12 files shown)


fpena (approved):
Looks good!

PR #1498, opened 2024-08-14
fix: BigInts on hermes android were crashing during serialization for reactotron

### LINKED ISSUE #1436: Reactotron does not support `BigInt()`
A community member raised an issue [in the reactotron channel](https://infiniteredcommunity.slack.com/archives/C42C5UKHV/p1707258028882889) that the library doesn't support `BigInt()`.

Indeed, it does not.

```tsx
<Button
  text="BigInt(0)"
  onPress={() => {
    console.log(BigInt(0))
  }}
/>
```

image

markrickert:

## Describe your PR

I though I had fixed this but we got more reports from the community that it was still broken on android.

I added a few new buttons to the example app and was able to reproduce the crash on my android emulator on expo 50 / rn 0.73.6.

With these changes, bigints show up properly in the reactotron app. Closes #1436 

33A78A6F-C83B-4CA6-BDA2-8F3F6E1A5B16-75409-00059A133DB30F84


ivannzme:
thanks a lot for the fix, try with your changes for my app. working fine

kevinvangelder (approved):

apps/example-app/babel.config.js
```diff
@@ -11,7 +11,8 @@ module.exports = function (api) {
   api.cache(true)
   return {
     assumptions: {
-      "setPublicClassFields": true,
+      setPublicClassFields: true,
+      privateFieldsAsSymbols: true,
```
frankcalise, line 15: What are these for?

lib/reactotron-core-client/src/serialize.ts
```diff
@@ -89,6 +93,8 @@ function serialize(source, proxyHack = false) {
           return value
         case "number":
           return value
+        case "bigint":
+          return value.toString()
```
lindboe, line 97: @markrickert do you have context on why `toString` works for BigInt but not `toJSON`? just a little confusing that we have to override toJSON but toString works.

frankcalise (commented):
Looks good, I'm just wondering if the babel properties were added when trying something else?

apps/example-app/babel.config.js
frankcalise: Ok I see it's for the following console warnings, maybe let's note that in a comment?

```
[transform-private-methods]: You are using the "loose: true" option and you are explicitly setting a value for the "setPublicClassFields" assumption. The "loose" option can cause incompatibilities with the other class features plugins, so it's recommended that you replace it with the following top-level option:
transform[stderr]: 	"assumptions": {
transform[stderr]: 		"setPublicClassFields": true,
transform[stderr]: 		"privateFieldsAsSymbols": true
transform[stderr]: 	}
```

lindboe: we spent a little time trying to track down what's going on with  babel and why we might even need to set exceptional config around assumptions / the class property transform, kind of came up empty. @morganick @joshuayoes is this something you have context on, since you were the last to look at this config?

morganick: These settings are from the initial warning that said to set it to this value. It did not appear to help reduce the number of warnings. I unfortunately was not able to get to the end of this 🐰 🕳️ . It also did not break the build. I have a feeling the issue is subtle and isn't *currently* causing issue that we're away of.

lindboe: Yeah it's probably fine to merge, I'm just wondering if we might be carrying config that doesn't matter. Something else to follow up on later.

frankcalise (approved):

PR #1509, opened 2024-10-03
fix(example-app): remove react-native-reanimated

frankcalise:

## Describe your PR
- Removes `react-native-reanimated` dependency from the Example App
- Simplifies the Example App for easier contributions (in the terms of, it's less of a headache to build, see Why)
- Tested building Android, switching between old and new arch without any hiccups or patching that @morganick has begun to push upstream

## Why
- It's been trouble in a monorepo setting
- We don't use it in the example app, it was just from Ignite spin up


morganick (approved):
LGTM 👍 and works as expected.

PR #1520, opened 2024-11-02
Fix: Handling console.log(null)

### LINKED ISSUE #1517: Logging null crashes the app UI
### Describe the bug

`console.tron.log(null)` results in the app UI crashing.

image

### Reactotron version

3.7.0

morganick:

- [ ] ~I have added tests for any new features, if relevant!~
- [ ] ~`README.md` (or relevant documentation) has been updated with your changes~

## Describe your PR

Fixes #1517 where calling `console.log` with `null` resulted in a blank screen. Now the comments around the function say that it only handles `string`, `object`, `number`, and `boolean`; however, the function takes a `message` of type `any`. Seems like we could narrow that down a bit. I've checked with all of the possible return values for `typeof` and documented that in the code.

The reason the bug existed is because `typeof null` results in `object`. The code -- nor many people-- were expecting that. Here's the reference to the docs as to why: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null

It appears that we do some things to not return functions as strings and we don't handle `Symbol`. Not sure if that's a requirement based on what the comments around the function say that it should accept as a message type.

⚠️ I'm not sure why the `getPreview` function returned just the raw message if we didn't discern what type it was. That seems like a bug. If someone has context as to why that was the case, I'd love to understand why.

## Screenshot
This is what it looks like now:

image

Here's another case where we didn't handle `null` correctly and now we do:

image

Happy Debugging 😎

### DIFF
--- a/lib/reactotron-core-ui/src/timelineCommands/LogCommand/index.tsx
+++ b/lib/reactotron-core-ui/src/timelineCommands/LogCommand/index.tsx
@@ -174,37 +174,41 @@ function getLevelName(level: string) {
 
 // function getPreview(message: string | object | boolean | number) {
 function getPreview(message: any) {
-  if (typeof message === "string") {
-    return message.substr(0, 500)
-  } else if (typeof message === "object") {
-    const firstValues = {}
+  // Special case: typeof null === "object"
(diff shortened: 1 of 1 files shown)


kevinvangelder (approved):

frankcalise (approved):

jamonholmgren:
That's a wild bug.

PR #1524, opened 2024-11-14
Removing react-native-flipper

morganick:


apps/reactotron-app/src/renderer/contexts/Standalone/useStandalone.ts
```diff
@@ -97,7 +97,7 @@ export function reducer(state: State, action: Action) {
           )
         }
 
-        // TODO: Figure out if we can stop having these dumb commands so early. Make core client only send once we actually have a client ID! (maybe won't work for flipper if it doesn't assign client id? Maybe that is how this broke?!?)
+        // TODO: Figure out if we can stop having these dumb commands so early. Make core client only send once we actually have a client ID!
```
morganick, line 100: I did not investigate why we have these commands so early. I only dropped the reference to flipper.

lib/reactotron-react-native/src/flipper-connection-manager.ts
```diff
@@ -1,80 +0,0 @@
-import { addPlugin } from "react-native-flipper"
-
-import ConnectionManager from "./connection-manager"
```
morganick, line 3: This was the only place this was used so I removed it as well.

frankcalise (approved):
LGTM 🚀

morganick:
@frankcalise found that the Android side of the example app wasn't building. The culprit was `react-native-screens`. Went ahead and upgraded that. Probably should do another upgrade pass here soon anyways.

Issue #1430, opened 2024-01-30
Bunch of TS errors from reactotron-core-client in a fresh project


EdwardDrapkin:
tsconfig:

```json

{
  "extends": "@react-native/typescript-config/tsconfig.json",
  "compilerOptions": {
    "baseUrl": "src",
    "rootDir": ".",
    "strict": true,
    "strictBindCallApply": true,
    "strictFunctionTypes": true,
    "strictNullChecks": true,
    "strictPropertyInitialization": true,
    "pretty": true,
    "allowImportingTsExtensions": false,
    "allowJs": false,
    "noImplicitAny": true,
    "noImplicitOverride": true,
    "noImplicitThis": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "noUncheckedIndexedAccess": true,
    "useUnknownInCatchVariables": true,
    "noEmit": true,
    "module": "ES2022",
    "typeRoots": ["./node_modules/@types", "./src/@types"],
  },
  "exclude": [
    "node_modules/**/*",
    "babel.config.js",
    "metro.config.js",
    "jest.config.js",
    "eslint.config.js",
    "config/**/*"
  ]
}

```

Include location:

```ts

import Reactotron from 'reactotron-react-native';

export const reactotron = Reactotron.configure({
  name: 'Mobile',
}) // controls connection & communication settings
  .useReactNative() // add all built-in react native plugins
  .connect(); // let's connect!
```

tsc output:

```
/node_modules/.bin/tsc
node_modules/reactotron-core-client/src/plugins/benchmark.ts:10:11 - error TS7034: Variable 'steps' implicitly has type 'any[]' in some locations where its type cannot be determined.

10     const steps = []
             ~~~~~

node_modules/reactotron-core-client/src/plugins/benchmark.ts:13:54 - error TS7005: Variable 'steps' implicitly has an 'any[]' type.

13       const previousTime = steps.length === 0 ? 0 : (steps[steps.length - 1] as any).time
                                                        ~~~~~

node_modules/reactotron-core-client/src/plugins/benchmark.ts:20:52 - error TS7005: Variable 'steps' implicitly has an 'any[]' type.

20       reactotron.send("benchmark.report", { title, steps })
                                                      ~~~~~

[...]

PR #1530, opened 2025-01-03
fix(package.json): remove "react-native" export condition

joshuayoes:

## Describe your PR
This PR removes the `"react-native"` export condition from all of our Reactotron packages. This is a shortcut fix for https://github.com/infinitered/reactotron/issues/1430

This PR is essentially the same solution as this community patch: https://github.com/infinitered/reactotron/issues/1430#issuecomment-1946318925

This was added in https://github.com/infinitered/reactotron/pull/1326 to improve errors and stack traces, but it appears that some issues with Reactotron's types are causing compilation issues with people's bundler and test setups. 

This should ultimately be fixed by making Reactotron's types more robust and strict, but to improve our user experience now, we are implementing this solution.

### DIFF
--- a/scripts/package.validate.mjs
+++ b/scripts/package.validate.mjs
@@ -1,47 +1,41 @@
 #!/usr/bin/env zx
 // @ts-check
-import "zx/globals";
-import { getWorkspaceList } from "./tools/workspace.mjs";
-$.verbose = false;
+import "zx/globals"
+import { getWorkspaceList } from "./tools/workspace.mjs"
+$.verbose = false
 
 console.log(
-  [
-    "Validating each workspace has the correct package.json fields for publishing.",
-  ].join("\n")
-);
+  ["Validating each workspace has the correct package.json fields for publishing."].join("\n")
+)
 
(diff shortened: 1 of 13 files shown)


scripts/package.validate.mjs
```diff
@@ -103,160 +85,131 @@ for (const workspacePath of workspacePaths) {
...
+  // assert "exports.react-native" field does not exist
+  if (packageJson.exports["react-native"]) {
+    console.error(
+      [
+        `Invalid ${packageJsonPath} exports.react-native field should not exist`,
+        `Remove this check once typescript types are more stable`,
+        `See https://github.com/infinitered/reactotron/issues/1430`,
+      ].join("\n")
+    )
+    process.exit(1)
```
joshuayoes, line 168: Prettier formatting made a bunch of noise in this file but this is the actual change

jamonholmgren (approved):
Looks good, thanks for jumping on this @joshuayoes !

markrickert (approved):

PR #1533, opened 2025-01-23
Export timeline log feature

denis-remitly:

## Describe your PR
Screenshot 2025-01-23 at 12 24 52

This PR adds the ability to export the timeline log into a JSON file.


jamonholmgren:
Very cool feature!

coolsoftwaretyler (changes requested):
Hey @denis-remitly - thanks for this! I really like it a lot. I have two suggestions. I'd love if you can tackle both in this PR before we merge, but only one of them is really blocking (I think an empty timeline should be an empty array, rather than nothing at all).

coolsoftwaretyler (approved):
I pushed up a commit to write empty arrays when there are no commands, and we can handle the Download path issue in a follow up. Running CI now and will get another IR review.

PR #1545, opened 2025-03-01
Removes storybook from codebase entirely, including storybook switch feature

jamonholmgren:
This is a rather large PR, but the actual changes are fairly limited in surface area. It's part of my new effort to streamline and simplify Reactotron's code base to make it easier to work on.

1. Removed Storybook on/off feature, as it's a confusing feature to set up, of minimal value, and we don't use Storybook anymore anyway
2. Removed Storybook entirely from the dev code base too
3. Fixed a bunch of dependency issues
4. Fixed a bunch of lint / formatting errors

To review, mostly just look at the Storybook commit, unless you like looking at linting changes.

- [x] Tested manually, works


.yarnrc.yml
```diff
@@ -10,3 +10,12 @@ npmAuthToken: "${NPM_TOKEN-}"
...
 yarnPath: .yarn/releases/yarn-4.1.1.cjs
+
+# Ignore specific warnings
+logFilters:
+  - code: YN0057 # Resolutions field will be ignored
+    level: discard
+  - code: YN0060 # Peer dependency issues
+    level: discard
+  - code: YN0086 # Some peer dependencies are incorrectly met
+    level: discard
```
frankcalise, line 21: any idea what this is from? I don't know enough about yarn 4 to know if it is right or wrong

frankcalise (commented):
Is it spring already? 🧹 

Generally looks fine to me, spotted one yarn config change that I'm unsure what it does

.yarnrc.yml
jamonholmgren: Cursor added it when I asked it to make yarn stop yelling at me about peer dependencies that didn’t seem to break anything. It just muted those warnings.

silasjmatson (approved):
Thanks for separating the lint/formatting fixes into their own commit. It made this much more reviewable. :shipit:

frankcalise (approved):

PR #1547, opened 2025-03-10
Feature/auto light theme

arkthur:

## Describe your PR

Added a light theme and logic that switches themes according to system preferences.


lib/reactotron-core-ui/src/theme.ts
```diff
@@ -1,4 +1,4 @@
-export default {
+const darkTheme = {
```
coolsoftwaretyler, line 1: **nitpick:** can we export both of these as named exports? It's strange to me that one theme is the default export and another gets named.

lib/reactotron-core-ui/src/components/TreeView/index.tsx
```diff
@@ -30,19 +30,29 @@ const MutedContainer = styled.span`
   color: ${(props) => props.theme.highlight};
 `
 
-const treeTheme = {
+const getTreeTheme = (baseTheme: Record<string, any>) => ({
```
coolsoftwaretyler, line 33: **issue:** if possible, I'd like us to write the types for a theme so that we can both avoid the `any`, and also make it possible for other people to get type checking if we want to write additional themes in the future.

Something like this, maybe:

```ts
interface ReactotronTheme {
  fontFamily: string;
  background: string;
  backgroundDarker: string;
  backgroundHighlight: string;
  backgroundLight: string;
  backgroundLighter: string;
  backgroundSubtleDark: string;
  backgroundSubtleLight: string;
  bold: string;
  chrome: string;
  chromeLine: string;
  constant: string;
  foreground: string;
  foregroundDark: string;
  foregroundLight: string;
  glow: string;
  heading: string;
  highlight: string;
  keyword: string;
  line: string;
  modalOverlay: string;
  string: string;
  subtleLine: string;
  support: string;
  tag: string;
  tagComplement: string;
  warning: string;
}
```

coolsoftwaretyler (changes requested):
Hey @arkthur - thanks for the PR! I had a couple small points of feedback for you, but other than that I see things working quite well.

https://github.com/user-attachments/assets/73a628ef-41b6-41b8-9adc-a4cae22fd5d0

lib/reactotron-core-ui/src/theme.ts
arkthur: I also found it weird, but it was the less disruptive way. Now I took it to the other extreme and renamed the file and put both themes in a const.

lib/reactotron-core-ui/src/components/TreeView/index.tsx
arkthur: Completely agree! I added it just like you suggested

arkthur:
Now I added a custom hook so the listener can be properly removed on unmount and the logic is less repetitive, avoiding redundancy.

arkthur:
Just pushed a fix for the unit tests, which includes some needed changes to the hook

frankcalise (commented):
This all looks good to me!

One last check though, @arkthur could you include a screenshot of the `<TreeView />` since there were changes there. Or a video of the timeline and expanding some JSON response for example.

Thanks!

arkthur:
@frankcalise as per requested
Screenshot 2025-03-12 at 12 37 05

arkthur:
One last thing, though.

https://github.com/user-attachments/assets/330dc604-c51f-461f-9879-f0aba6589841

Looking into it right now.

arkthur:
@frankcalise @coolsoftwaretyler Now we're talking!

https://github.com/user-attachments/assets/c05dcdd8-f2bf-4fc7-9ed1-18e8fd0ff52c

frankcalise (approved):
Awesome work, thanks for the extra thorough checks

PR #1613, opened 2026-06-09
feat(reactotron-react-native): track Expo expo/fetch in the networking plugin

### LINKED ISSUE #1612: Network requests not captured on Expo SDK 56
### Summary

On Expo SDK 56, Reactotron's networking timeline shows **no network requests**. SDK 56 installs `expo/fetch` as the global `fetch` by default, and `expo/fetch` is a native implementation that does **not** go through `XMLHttpRequest`. Reactotron's network instrumentation only patches `XMLHttpRequest`, so it never observes any traffic. Console `log`/`display` events are unaffected — only the network timeline is empty.

This affects **both** project types:

- **Managed Expo apps** — `expo` is loaded at startup (e.g. via `expo-router/entry`), so the `fetch` swap happens from app launch and the network timeline is empty from the start.
- **Bare React Native apps that use Expo packages** — the swap runs the first time anything imports the `expo` package at runtime. Importing an Expo package triggers `expo/src/Expo.fx`, which loads the Winter runtime and replaces `global.fetch`. For example, calling `useCameraPermissions()` from `expo-camera` (which imports from `expo`) is enough. So in a bare app the network timeline can appear to work and then go empty once such a code path is loaded. (A type-only reference like `useRef<CameraView>(null)` is elided by the Babel pipeline and does NOT trigger it — only a runtime import does.)

### Environment

- `reactotron-react-native`: 5.2.0 (also confirmed against `master`, last pushed 2026-05-28)
- Reactotron desktop: 3.11.0
- Expo SDK: 56
- React Native: 0.85.3 (New Architecture)
- Platform: iOS & Android
- Reproduced on **both** a managed Expo app and a bare RN app that uses Expo
  packages (hybrid bare workflow)

### Root cause

Expo SDK 56 makes `expo/fetch` the default `globalThis.fetch` ([Expo docs](https://docs.expo.dev/versions/latest/sdk/expo/), [SDK 56 changelog](https://expo.dev/changelog/sdk-56-beta)). It's a fully native fetch that bypasses `XMLHttpRequest`.

Reactotron's interceptor patches **only** XHR:

- `lib/reactotron-react-native/src/xhr-interceptor.ts` — patches `XMLHttpRequest.prototype.open` / `send` / `setRequestHeader`
- `lib/reactotron-react-native/src/plugins/networking.ts` — relies solely on `XHRInterceptor.enableInterception()`

The assumption is even documented in `xhr-interceptor.ts`:

> "This supports interception with XMLHttpRequest API, including Fetch API and any other third party libraries that depend on XMLHttpRequest."

That assumption — that `fetch` is built on XHR — no longer holds under SDK 56's `expo/fetch`.

### Steps to reproduce

1. An SDK 56 app — either a managed Expo app, or a bare RN app that imports an  Expo package at runtime (default config, so `expo/fetch` is the global`fetch`).
2. Configure Reactotron with `.useReactNative()` (networking enabled) and `.connect()`.
3. Make any `fetch(...)` request (in a bare app, ensure the `expo` package has been imported — e.g. `useCameraPermissions()` from `expo-camera`).
4. Observe the Reactotron timeline.

### Expected

The request appears in the networking timeline.

### Actual

Network logs are missing in Timeline. App stays connected.

### Workaround

  Until the fix lands, add this line to your `.env` file:

  ```
  EXPO_PUBLIC_USE_RN_FETCH=1
  ```

  This is Expo's own opt-out — it keeps React Native's XHR-based `fetch` as the global, which Reactotron's existing networking plugin already instruments, so requests show up in the timeline again.

  Trade-off: the app forgoes `expo/fetch`'s improvements (brotli/gzip/zstd decompression, `AbortSignal.timeout`). Because `EXPO_PUBLIC_*` vars are inlined into the JS bundle at build time, prefer a dev-only env file (e.g.`.env.development`) so the opt-out doesn't ship to production.

### Related / prior art

Sentry hit and **fixed** this exact problem:

- Issue: getsentry/sentry-react-native#6225
- Fix PR: getsentry/sentry-react-native#6226

Their approach: detect that the global `fetch` is the native Expo implementation via the marker `globalThis.fetch[Symbol.for('expo.builtin')] === true` (a new `isExpoFetchEnabled()` helper), then enable fetch-layer instrumentation when detected (they flipped `traceFetch` and fetch-breadcrumb defaults from `false` to `isExpoFetchEnabled()`).

### Possible fix

**The catch for Reactotron:** Sentry already *had* a fetch-instrumentation layer and only needed to change a default. Reactotron is XHR-only (`xhr-interceptor.ts` / `networking.ts` patch `XMLHttpRequest` exclusively), so there's no existing fetch layer to enable — a fetch interceptor would need to be **added**.

A minimal version: wrap `globalThis.fetch` when the `expo.builtin` marker is present and emit `reactotron.apiResponse(request, response, duration)` — the same event the XHR path produces. Gating on the marker avoids double-counting when `EXPO_PUBLIC_USE_RN_FETCH=1` reverts the global to RN's XHR-based fetch.

### Reactotron version

5.2.0

ShanavasPS:

## Describe your PR

  Closes #1612 .

### Problem

Expo SDK 56 installs `expo/fetch` as the default `globalThis.fetch`. `expo/fetch` is backed by a native module and **bypasses `XMLHttpRequest`**, so the `networking` plugin's `XHRInterceptor` never sees it — on Expo SDK 56+ the plugin silently misses all `fetch` traffic.

(For context, the existing `XHRInterceptor` only covers React Native's XHR-backed `fetch`. The same blind spot affects other XHR-based network inspectors on Expo SDK 56+.)

### Solution

Add a `FetchInterceptor` that mirrors `XHRInterceptor`'s shape (`set*Callback` / `enableInterception` / `disableInterception`) and wraps the global `fetch` **only when it is the `expo/fetch` builtin**, detected via `Symbol.for("expo.builtin")`. React Native's XHR-backed `fetch` is left to `XHRInterceptor`, so there is **no double-reporting**. The `networking` plugin wires Reactotron into it in `onConnect`, gated by a new `ignoreExpoFetch` option.

Care was taken to keep this safe for `expo/fetch`'s streaming model:

- The wrapper returns the original `Response` **immediately** and reads the body off a **clone**, asynchronously — the caller is never blocked and streaming bodies stay intact.
- `text/event-stream` (and the existing image content types) are **never cloned/read**, so streaming responses aren't buffered.
- Request URL/method/header parsing for the `Request`/`URL`/string argument forms is handled in the interceptor; query-param parsing, content-type policy, and timing stay in the plugin (matching the XHR path).

### Files

- `lib/reactotron-react-native/src/fetch-interceptor.ts` — new, mechanism only (mirrors `xhr-interceptor.ts`)
- `lib/reactotron-react-native/src/fetch-interceptor.test.ts` — unit tests (no-op off Expo, wrap/callbacks, rejection, no double-wrap, restore on disable)
- `lib/reactotron-react-native/src/plugins/networking.ts` — wire in `FetchInterceptor`, add `ignoreExpoFetch` option
- `docs/plugins/networking.md` — document expo/fetch support and `ignoreExpoFetch`

### Notes / known limitations

- Detection is by symbol, so it is a no-op on non-Expo runtimes and on SDKs where `globalThis.fetch` is not `expo/fetch`.
- Only `globalThis.fetch` is wrapped. Code that does `import { fetch } from 'expo/fetch'` and captures the reference directly bypasses this (as it does the XHR interceptor); the global wrap covers the common case.
- Interception is installed in `onConnect`, so requests fired before Reactotron connects aren't captured.

### DIFF
--- a/docs/plugins/networking.md
+++ b/docs/plugins/networking.md
@@ -6,6 +6,8 @@ title: Networking
 
 The `networking` plugin is `reactotron-react-native` which allows you to track all XMLHttpRequests in React Native.
 
+On Expo SDK 56+, `expo/fetch` is installed as the default `globalThis.fetch`. It is backed by a native module and bypasses `XMLHttpRequest`, so it would otherwise be invisible to this plugin. The plugin detects this case and tracks `expo/fetch` requests as well (see `ignoreExpoFetch` below to opt out). Requests made through React Native's regular XHR-backed `fetch` continue to be tracked via `XMLHttpRequest`, so there is no double-reporting.
+
 ## Usage
 
 To use the `networking` plugin, you need to add the additional plugin on the `import` line.
@@ -26,10 +28,11 @@ And you're done! Now you can see your XMLHttpRequests in Reactotron.
 
 ## Advanced Usage
 
-`networking()` also accepts an object with two options:
+`networking()` also accepts an object with these options:
 
-- `ignoreContentTypes`: a regular expression which, when matched against the `Content-Type` response header, will prevent the data from being displayed in Reactotron. You typically want to do this for images (which is the default).
-- `ignoreUrls`: a regular expression which, when matched against the URL of the XHR, will prevent the request from being tracked in Reactotron. Can be useful for ignoring noisy logging requests.
+- `ignoreContentTypes`: a regular expression which, when matched against the `Content-Type` response header, will prevent the data from being displayed in Reactotron. You typically want to do this for images (which is the default). `text/event-stream` response bodies are always skipped so streaming responses are not buffered.
+- `ignoreUrls`: a regular expression which, when matched against the URL of the request, will prevent the request from being tracked in Reactotron. Can be useful for ignoring noisy logging requests.
+- `ignoreExpoFetch`: set to `true` to skip instrumenting Expo's `expo/fetch` (the default `globalThis.fetch` on Expo SDK 56+). Has no effect on non-Expo runtimes, where the global fetch is XHR-backed and already covered by XHR tracking.
 
 ```js
 networking({
--- a/lib/reactotron-react-native/src/fetch-interceptor.test.ts
+++ b/lib/reactotron-react-native/src/fetch-interceptor.test.ts
@@ -101,6 +101,37 @@ describe("FetchInterceptor", () => {
     expect(globalThis.fetch).toBe(wrapped)
   })
 
+  it("wraps an explicitly provided fetch even without the expo symbol", async () => {
+    const response = makeResponse(200, { "content-type": "application/json" })
+    // no expo.builtin symbol — e.g. expo-router's re-wrapped global fetch
+    const routerWrapped: any = jest.fn(() => Promise.resolve(response))
+    globalThis.fetch = routerWrapped
+
+    const open = jest.fn()
+    FetchInterceptor.setOpenCallback(open)
+    FetchInterceptor.enableInterception(routerWrapped)
+
+    expect(FetchInterceptor.isInterceptorEnabled()).toBe(true)
+    expect(globalThis.fetch).not.toBe(routerWrapped)
+    // the wrapper must not pretend to be the expo builtin
+    expect((globalThis.fetch as any)[EXPO_BUILTIN]).toBeUndefined()
+
+    const result = await (globalThis.fetch as any)("https://example.com/y")
+    expect(result).toBe(response)
+    expect(routerWrapped).toHaveBeenCalledTimes(1)
+    expect(open).toHaveBeenCalledTimes(1)
+    expect(open.mock.calls[0][1]).toBe("https://example.com/y")
+  })
+
+  it("does not wrap an explicitly provided fetch twice", () => {
+    const fn: any = jest.fn()
+    globalThis.fetch = fn
+    FetchInterceptor.enableInterception(fn)
+    const wrapped = globalThis.fetch
+    FetchInterceptor.enableInterception(fn)
+    expect(globalThis.fetch).toBe(wrapped)
+  })
+
   it("restores the original fetch on disable", () => {
     const original = makeExpoFetch(() => Promise.resolve(makeResponse(200, {})))
     globalThis.fetch = original
--- a/lib/reactotron-react-native/src/fetch-interceptor.test.ts
+++ b/lib/reactotron-react-native/src/fetch-interceptor.test.ts
@@ -1,0 +1,113 @@
+import { FetchInterceptor } from "./fetch-interceptor"
+
+const EXPO_BUILTIN = Symbol.for("expo.builtin")
+
+function makeExpoFetch(impl: (...args: any[]) => Promise<any>) {
+  const fn: any = (...args: any[]) => impl(...args)
+  fn[EXPO_BUILTIN] = true
+  return fn
+}
+
+function makeResponse(status: number, headersObj: Record<string, string>) {
+  return {
+    status,
+    headers: {
+      get: (k: string) => headersObj[k.toLowerCase()] ?? null,
+      forEach: (cb: (v: string, k: string) => void) =>
+        Object.entries(headersObj).forEach(([k, v]) => cb(v, k)),
+    },
+    clone() {
+      return this
+    },
+    text: () => Promise.resolve(""),
+  }
+}
+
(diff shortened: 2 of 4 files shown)


joshuayoes (changes requested):
Thanks for the thorough PR — the interceptor design is sound, and we verified it at runtime (Expo SDK 57 app on iOS simulator → Reactotron desktop built from this branch): when `globalThis.fetch` carries `Symbol.for("expo.builtin")`, the fetch traffic shows up in the timeline with correct method/URL/query params/status/timing/body, the caller's response is untouched, and there's no double-reporting alongside the XHR interceptor. Unit tests, lint, and typecheck all pass locally. Your claims about the expo internals (symbol stamping in `installGlobal`, native URLSession/OkHttp backing, lazy stamping on first global read, `clone()` tee semantics) all check out against the expo source.

**However, runtime testing shows the fix silently no-ops on expo-router apps — the default `create-expo-app` template and the primary managed-app scenario in #1612.**

Root cause (confirmed in both `sdk-56` and `sdk-57` branches of expo/expo): `expo-router/entry` imports `@expo/metro-runtime`, whose `src/location/install.native.ts` re-wraps the global fetch via `wrapFetchWithWindowLocation` and reassigns it with `Object.defineProperty(global, 'fetch', …)`. The wrapper function only carries `Symbol.for("expo.polyfillFetchWithWindowLocation")` — the `expo.builtin` stamp is lost — so `FetchInterceptor.enableInterception()` bails at the symbol check. This wrap is unconditional in dev **and** prod unless the app sets `expo.extra.router.origin: false`. We reproduced this end-to-end: with this PR's build installed in an expo-router SDK 57 app, `fetch` requests still never reach the timeline (the XHR control does). Bare RN apps using expo packages are unaffected, so the fix does work there.

Suggestion: detection needs to also handle the router wrapper — but keying on the polyfill symbol alone would be wrong, because `wrapFetchWithWindowLocation` also wraps the XHR-backed fetch when `EXPO_PUBLIC_USE_RN_FETCH=1`, which would double-report. Options worth exploring: unwrap/inspect the inner function, or compare against `require('expo/fetch').fetch` identity when the module is resolvable.

Two smaller, non-blocking notes:

- On expo 56.0.0–56.0.7 there's a known `bodyUsed` bug when a response is cloned twice (expo/expo#46397, fixed in 56.0.8). Reactotron's `clone()` plus user code's own `clone()` can trigger it — worth a caveat in `docs/plugins/networking.md`.
- `onFetchOpen` runs synchronously inside the caller's `fetch()` call, and the query-param `decodeURIComponent` can throw on malformed input (e.g. `?q=%zz`), which would break the app's request. Worth a try/catch around the param parsing.

Happy to re-verify once the router case is covered — the runtime harness is set up and quick to re-run.

joshuayoes:
Following up on the expo-router gap above with a concrete proposal: ShanavasPS/reactotron#1 (stacked on this PR's branch, for you to verify and merge into this PR if you like the approach).

Instead of more environment sniffing, it uses dependency injection — the app passes the exact fetch reference it wants tracked:

```ts
.useReactNative({ networking: { fetch: globalThis.fetch } })
```

Auto-detection stays the default so nothing changes for existing users, but expo-router apps get a one-line fix instead of a silent no-op — and injecting a reference composes with other non-XHR fetch implementations too (e.g. [react-native-nitro-fetch](https://github.com/margelo/react-native-nitro-fetch); untested there, only the expo path was verified). We verified it at runtime on an expo-router SDK 57 iOS app where this branch currently no-ops: with the option set, fetch traffic appears in the timeline with correct method/params/status, the caller's response is untouched, and XHR reports exactly once.

joshuayoes (approved):
Approving — the blocker from my earlier review (silent no-op on expo-router apps, where the router's location polyfill drops the `expo.builtin` symbol) now has a workable escape hatch in this PR: the explicit `fetch` option (`networking: { fetch: globalThis.fetch }`), which I verified at runtime on an expo-router SDK 57 app on the iOS simulator.

The remaining review items (error isolation so instrumentation can never break the app's own `fetch`, good-citizen restore on disable, safe `clone()`, and docs for the `fetch` option) are implemented with tests in the stacked follow-up #1615, which should land right after this. Automatic detection for expo-router apps is the next iteration, tracked in #1612.

CI is green on this head via the trusted-branch run (build_and_test, build_docs). Thanks @ShanavasPS for the solid base fix!

